PR #21, opened 2026-03-27
feat(TimezoneWidget): add option for custom timezones

Akashic101:
With this PR a user can create their own custom timezone
image


mauriceboe:
Hey @Akashic101, built this in natively. You can now add custom timezones with any valid IANA timezone identifier (e.g. America/New_York, Asia/Kolkata). Optional label, validation, duplicate check. All three languages. Thanks for the idea!

PR #22, opened 2026-03-27
feat(PlaceFormModal): automatically split up latitude and longitude on paste

Akashic101:
When pasting a latitude longitude value such as 52.85702181513319, 11.828981343314497 it automatically gets split up into the correct inputs. Pasting just one value does not change anything


mauriceboe:
Hey @Akashic101, good idea! Built this in natively. Pasting "52.857, 11.829" into the latitude field now auto-splits into lat and lng. Works with comma, semicolon, or space as separator.

PR #30, opened 2026-03-27
Add right-click context menu to map view for adding and finding nearby places

Tiboriuss:
Adds a right-click (context menu) interaction to the trip planner map. Right-clicking anywhere on the map — or on an existing place marker — opens a two-item menu:

Add Place to List — opens the place form pre-filled with the clicked coordinates
Show Close Places — opens a bottom panel to discover nearby POIs
New: Close Places panel
The "Show Close Places" panel matches the existing PlaceInspector design (frosted glass, bottom-anchored, same header/footer layout) and contains:

Search radius slider — steps from 50 m up to 1 km, defaulting to 200 m
Search button — queries only when explicitly triggered (no auto-fire on open)
Radius circle on the map — a dashed indigo circle is drawn on the map that updates live as you drag the slider, so you can see the search area before searching
"In your trip list" section — shows existing trip places already within 1 km of the clicked point, sorted by distance
"Nearby" section — fetches named POIs from the free Overpass API (OpenStreetMap data, no API key required) — covers amenities, tourism, historic, leisure, and shop nodes
Smart address fill — when adding a POI, the address is extracted directly from the OSM tags (addr:street, addr:housenumber, addr:city, addr:postcode) or falls back to a Nominatim reverse-geocode call, so the place form opens with coordinates and address already filled in
"Add custom place here" button as a fallback for any location without a matching POI

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,6 +15,8 @@ client/public/icons/*.png
 # User data
 server/data/
 server/uploads/
+data/
+uploads/
 
 # Environment
 .env
--- a/client/src/components/Map/MapView.jsx
+++ b/client/src/components/Map/MapView.jsx
@@ -1,6 +1,6 @@
 import React, { useEffect, useRef, useState, useMemo } from 'react'
 import ReactDOM from 'react-dom'
-import { MapContainer, TileLayer, Marker, Tooltip, Polyline, useMap } from 'react-leaflet'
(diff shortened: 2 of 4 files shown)


mauriceboe:
Hey @Tiboriuss, thanks for the effort! We actually built right-click to add places natively in 2.6.2 with auto reverse geocoding. Your nearby places panel with the Overpass radius search is a cool idea though, might revisit that concept later. Closing this for now since the core feature is in.

PR #66, opened 2026-03-28
feat: add OIDC-only mode to disable password authentication

### LINKED ISSUE #62: [FEATURE] Allow disabling of password auth when OIDC is enabled
**Is your feature request related to a problem? Please describe.**
Even when OIDC auth is enabled, it is still possible to login with just email and password. This is not ideal if you want to make the app public-facing so that others can access it, but still be protected from password stuffing or brute-force attempts. Being able to disable password auth and rely solely on the OIDC identity provider would be a much better experience.

**Describe the solution you'd like**
The ability to disable password auth when OIDC is enabled. Ideal, logins should just be forwarded straight to the identity provider

**Describe alternatives you've considered**
Using a reverse-proxy to disable the password auth routes, but this isn't ideal.

**Additional context**
Password stuffing and other password-based attacks can only really be prevented by requiring MFA and disabling any ability to auth with just a username and password.

swheettaos:
This is my attempt to close #62 and it also fixes a couple things I found with TS express versioning and schduler.ts for the cron

## Summary

When OIDC/SSO is configured, admins can now enable a "Disable password
authentication" toggle in Admin → Settings → SSO. This addresses the
concern raised in #62: even with OIDC enabled, the app previously allowed
login via email/password, leaving it exposed to password stuffing and
brute-force attacks. With this change, admins can force all authentication
through the configured identity provider and completely block password-based
login, registration, and password changes.

### Backend
- **`routes/auth.ts`**: Added `isOidcOnlyMode()` helper; blocks `POST /auth/login`,
  `POST /auth/register` (for non-first-user), and `PUT /auth/me/password`
  with HTTP 403 when OIDC-only mode is active
- **`routes/auth.ts`**: Exposes `oidc_only_mode` boolean in `GET /auth/app-config`
- **`routes/admin.ts`**: Exposes `oidc_only` flag on `GET /admin/oidc` and accepts
  it on `PUT /admin/oidc` (persisted to `app_settings`)

### Frontend
- **`AdminPage`**: Toggle in the OIDC/SSO settings section (`oidc_only` saved
  with the rest of OIDC config on the same Save button)
- **`LoginPage`**: When `oidc_only_mode` is active, replaces the login form
  with a single OIDC redirect button and hides the register toggle
- **`SettingsPage`**: Hides the password change section when `oidc_only_mode` is on
- **i18n (en/de)**: Added keys `admin.oidcOnlyMode`, `admin.oidcOnlyModeHint`,
  `login.oidcOnly`

### TypeScript / Build fixes
These were pre-existing latent errors made visible when `node_modules` were
first installed locally. Included here as they must be resolved for the
project to compile cleanly.

**Root cause:** `@types/express@^5.0.6` was listed in `devDependencies` while
the runtime is `express@^4.18.3`. The v5 type definitions widened
`ParamsDictionary` from `string` to `string | string[]`, causing 115 type
errors across every route handler that accessed `req.params`.

**Fix:** Downgraded `@types/express` to `^4.17.25` (latest v4) to match the
runtime. `req.params` is correctly typed as `string` again — no workarounds
or casts needed.

**`scheduler.ts`:** `cron.ScheduledTask` was referenced as a namespace-qualified
type, but the installed version of `node-cron` does not export `ScheduledTask`
as a property of the default export. Fixed by using a named type import:
```ts
import cron, { type ScheduledTask } from 'node-cron';
```

## Behaviour

| Scenario | UI behaviour | Direct API call |
|---|---|---|
| OIDC configured, `oidc_only = false` (default) | No change — password auth still works | No change |
| OIDC configured, `oidc_only = true` | Login form replaced with SSO button; register + password change sections hidden | 403 returned |
| OIDC not configured | Toggle has no effect | — |
| First-user registration | Always shown | Always allowed |

### DIFF
--- a/server/package-lock.json
+++ b/server/package-lock.json
@@ -30,7 +30,7 @@
         "@types/bcryptjs": "^2.4.6",
         "@types/better-sqlite3": "^7.6.13",
         "@types/cors": "^2.8.19",
-        "@types/express": "^5.0.6",
+        "@types/express": "^4.17.25",
         "@types/jsonwebtoken": "^9.0.10",
         "@types/multer": "^2.1.0",
         "@types/node": "^25.5.0",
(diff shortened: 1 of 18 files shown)


swheettaos:
I'd test this a bit .. I had some AI help in the tscript stuff.. it's not my strongest area .. feel free to reject if I messed stuff up.

PR #69, opened 2026-03-29
fix: resolve static asset SSL errors from helmet's upgrade-insecure-requests

saswatds:
Helmet merges default CSP directives (including `upgrade-insecure-requests`) into custom directives when `useDefaults` is true (the default). This caused browsers to upgrade all HTTP sub-resource requests to HTTPS, breaking static assets when the server runs over plain HTTP.

This commit conditionally sets `upgrade-insecure-requests` based on FORCE_HTTPS: enabled in production (where HTTPS is available), explicitly disabled (null) otherwise to prevent browser SSL errors on home servers and development environments.

Also extracts `shouldForceHttps` to avoid repeated env lookups.

| Before | After |
|--------|--------|
|image | image|

### DIFF
--- a/server/src/index.ts
+++ b/server/src/index.ts
@@ -44,6 +44,8 @@ if (allowedOrigins) {
   corsOrigin = true;
 }
 
+const shouldForceHttps = process.env.FORCE_HTTPS === 'true';
+
 app.use(cors({
   origin: corsOrigin,
   credentials: true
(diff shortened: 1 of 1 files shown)


mauriceboe (approved):
Good catch, this was the missing piece. We fixed HSTS and the redirect but missed that Helmet's default CSP still included upgrade-insecure-requests. Thanks\!

PR #72, opened 2026-03-29
fix: prioritize ADM0_A3 over ISO_A3 in atlas area resolution to support France, Norway and Israel

Summerfeeling:
I noticed that France isn't displayed in the atlas when you've been there. As far as I can tell, this is because `feature.properties?.ISO_A3` returns -99 as the country point at this point instead of FRA. `feature.properties?.ADM0_A3` returns the correct value.

While testing, I noticed that the same issue existed for Norway and Israel. These countries are now displayed as well. 

For all other countries, ISO_A3 and ADM0_A3 match. I have therefore prioritized ADM0_A3 over ISO_A3 to fix the error with these three countries.

### DIFF
--- a/client/src/pages/AtlasPage.tsx
+++ b/client/src/pages/AtlasPage.tsx
@@ -241,7 +241,7 @@ export default function AtlasPage(): React.ReactElement {
       interactive: true,
       bubblingMouseEvents: false,
       style: (feature) => {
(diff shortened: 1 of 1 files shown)


mauriceboe:
Great catch, thanks for the fix! 🎯

PR #80, opened 2026-03-29
feat(client): add time of day groupings

errnerr:
Started to use this to plan out our trip to Japan next month. I think I really useful feature would be to group planed activities by morning / afternoon / evening. 

### Behavior
If nothing on that day has a time, the list stays in the existing manual order (same as before).

If anything has a time, items are ordered as:

- No time – places/notes without a usable time stay at the top, in their previous order.
- Morning – 5:00–11:59
- Afternoon – 12:00–16:59
- Evening – everything else (17:00–23:59 and 0:00–4:59)

Within each band, items are sorted by clock time, then by the existing order / sort fields.

### Where times come from
- **Places**:` place.place_time `(HH:MM). If that’s empty, the linked reservation reservation_time (ISO with T) is used, same idea as the rest of the sidebar.
- **Notes**: first `HH:MM` match in the note’s time text (e.g. Meet 14:30 → 14:30).
UI

When at least one item on the day is timed, Morning / Afternoon / Evening labels (translated) are shown when the list moves into that band.

I've also added this to the PDF summary too.

Additionally, I've created a .devcontainer config which might be useful.

image

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -1,0 +1,6 @@
+FROM mcr.microsoft.com/devcontainers/javascript-node:22-bookworm
+
+# better-sqlite3 and other native npm modules
+RUN apt-get update && export DEBIAN_FRONTEND=noninteractive \
+    && apt-get install -y --no-install-recommends build-essential python3 \
+    && apt-get clean && rm -rf /var/lib/apt/lists/*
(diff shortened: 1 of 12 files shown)


mauriceboe:
Hey @errnerr! Thanks for the idea, but we're not going to merge this.

The upcoming release introduces automatic chronological sorting — places with times and transport bookings are now sorted by time in the day plan. This gives you a natural time-based flow without rigid morning/afternoon/evening groupings that might not fit every travel style.

If you need visual structure, day notes can be used as section dividers between parts of your day.

Closing this PR — appreciate the effort though!

Issue #45, opened 2026-03-28
[FEATURE] pre-trip checklist


oratory:
**Is your feature request related to a problem? Please describe.**

A lot of trips require taking care of things before leaving, like arranging dog sitters, tourist visas, prescription medicine, etc. 

**Describe the solution you'd like**

A simple checklist to tick off items as they are completed. 

**Describe alternatives you've considered**

A generic markdown doc in the files section I guess would work.

blitzmikyu:
I feel like the packing list section fills this need. Maybe it should be named pre-flight or any other alternative suitable title for the section that is more broad.

mauriceboe:
Hey @oratory, the packing list addon actually covers this already. Despite the name, it's really just a categorized checklist with progress tracking. You can create custom categories like "Before Trip" or "To Do" and add items like "arrange dog sitter", "apply for visa", "refill prescriptions" etc. Each item has a checkbox to tick off.

That said, renaming it to something more generic like "Checklists" or "Lists" could make that clearer. Good point.

PR #116, opened 2026-03-29
feat: native MCP addon — 35 tools, HTTP/SSE transport, per-user service tokens

swheettaos:
> This PR targets `dev`. This is a first attempt at a native MCP integration — happy to make changes based on your feedback before it goes anywhere near `main`.

---

Adds a native Model Context Protocol (MCP) addon that lets AI assistants (Claude, Cursor, etc.) read and manage TREK data via 35 tools.

## What's included

**New addon: MCP Server**
- Disabled by default; admin enables via Admin → Addons
- HTTP/SSE transport at `/api/mcp` using `@modelcontextprotocol/sdk`

**35 tools across 6 groups**
| Group | Tools |
|-------|-------|
| Trips | list, get, get_context, create, update, archive, delete, duplicate |
| Days | list, get, create, update, delete |
| Places | list, get, create, update, delete, assign_to_day, unassign_from_day |
| Packing | list_items, create_item, toggle, delete, list_categories, set_assignee |
| Budget | list, get_summary, create, update, delete |
| Search | search (cross-entity) |

**Tools call existing API routes (not the DB directly)**
- WebSocket broadcasts fire automatically — collaborators see AI changes in real time
- Permissions evaluated live from existing route logic — zero duplication

**Per-user service tokens**
- Any user creates their own token in Settings → AI / MCP
- Admin can audit and revoke all tokens in Admin → Addons → MCP Server
- Tokens are scoped to the creating user's permissions

**DB migration**
- Migration #45: `service_tokens` table
- MCP addon seeded (disabled by default)

### DIFF
--- a/README.md
+++ b/README.md
@@ -73,6 +73,7 @@
 - **Vacay** — Personal vacation day planner with calendar view, public holidays (100+ countries), company holidays, user fusion with live sync, and carry-over tracking
 - **Atlas** — Interactive world map with visited countries, bucket list with planned travel dates, travel stats, continent breakdown, streak tracking, and liquid glass UI effects
 - **Collab** — Chat with your group, share notes, create polls, and track who's signed up for each day's activities
+- **MCP Server** — Expose TREK to AI assistants (Claude, Cursor, etc.) via the Model Context Protocol with 35 tools covering trips, days, places, budgets, and packing
 - **Dashboard Widgets** — Currency converter and timezone clock, toggleable per user
 
 ### Customization & Admin
@@ -90,7 +91,8 @@
 - **PWA**: vite-plugin-pwa + Workbox
 - **Real-Time**: WebSocket (`ws`)
 - **State**: Zustand
-- **Auth**: JWT + OIDC + TOTP (MFA)
+- **Auth**: JWT + OIDC + TOTP (MFA) + MCP service tokens
+- **MCP**: `@modelcontextprotocol/sdk` (HTTP/SSE transport)
 - **Maps**: Leaflet + react-leaflet-cluster + Google Places API (optional)
 - **Weather**: Open-Meteo API (free, no key required)
 - **Icons**: lucide-react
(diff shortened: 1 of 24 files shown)


jubnl:
Was working on an MCP integration too and saw your pr. A proper dev should review and test AI generated code.

1. The auth middleware reads decoded.id. You sign the token with sub: user.id. So everything under /api/mcp will return a 401.
2. the JWT_SECRET is not always defined as an environment variable. Read `server/src/config.ts` and check the usage of the exported variable.
3. server/src/mcp/transport.ts authenticates the caller on POST /message but does not verify that the authenticated user owns the target session.
4. Service tokens grant the same access as a full session JWT, including settings changes, token creation, admin endpoints.
5. server/src/mcp/transports.ts again: `sessions` variable is a Map with no size cap, no TTL, and cleanup only on close.

And I found at least 4 other big issues.

Sorry if that's harsh but I'm tired of this kind of vibe code.

swheettaos:
This was my inital stab at it .. I submitted this to the dev branch .. I was more interested in the approach. Where it was a plugin, leveraged the existing api calls, had per user tokens (even if I didn't do it the right way) , had admin screen for token management.  I didn't say this was complete. In fact I said ..   "This PR targets dev. This is a first attempt at a native MCP integration — happy to make changes based on your feedback before it goes anywhere near main."  

But hey.. if my "vibe" coding sucks to bad then I let the experts do it and stop wasting my time.  I'll happily use the MCP you or someone else submits instead of my POS.  As you can probally tell, this is the first time I have ever tried to create a MCP and yes I did leverage AI to help me but it was by no means VIBE coding. It was just straight lack of knowledge.  I shouldn't have submitted it as a PR (even to the dev branch).   This isn't a programming class. 

I'll just keep to just submitting feature requests instead of trying to driectly contribute because it seems like he's more than enough people contributing now witth clearly better skills than me.

PR #117, opened 2026-03-29
fix: pass invite token through OIDC flow to allow invited registration

slashwarm:
invite links don't work when you try to register via OIDC/SSO - the token just gets lost in the redirect. so if you have registration disabled (which most people do), invited users can't sign in with google/etc.

this passes the invite token through the oidc state so it survives the round trip. valid invite bypasses the registration check, usage count gets bumped after.


mauriceboe:
Merged to `dev`, shipping today. Clean fix — passing the invite token through OIDC state is the right approach. Thanks @slashwarm!

PR #119, opened 2026-03-29
feat: add Hungarian language support

slashwarm:
adds Magyar (hu) as the 10th language. complete translations for all 1160+ keys including the new memories/immich stuff.


mauriceboe:
Merged to `dev`, added ~104 missing keys to match en.ts. Thanks @slashwarm for another contribution! Shipping today.

PR #120, opened 2026-03-29
feat: add leftWidth/rightWidth layout props to DayDetailPanel

slashwarm:
optional props so the parent can pass sidebar widths for centering the panel between the left sidebar and right panel. defaults to 0 so it's a no-op if not used.


mauriceboe:
Merged to `dev`, shipping today. Nice improvement — panels now center properly between the sidebars. Thanks @slashwarm!

PR #122, opened 2026-03-29
refactor: replace hardcoded date arrays with Intl.DateTimeFormat in VacayMonthCard

slashwarm:
VacayMonthCard has 12 hardcoded arrays (WEEKDAYS_EN/DE/ES/FR/BR/AR + MONTHS_EN/DE/ES/FR/BR/AR) and a ternary chain that falls back to English for any language it doesn't know about, so ru, zh, nl all show English month/day names.

replaced the whole thing with Intl.DateTimeFormat using the locale from useTranslation(). works for every locale out of the box, no maintenance needed when new languages are added.


mauriceboe:
Hey @slashwarm! Great refactor — applied your approach manually to `dev` since there were merge conflicts with other i18n changes. Credit in the commit message. Thanks for the cleanup, shipping today!

PR #125, opened 2026-03-30
feature: MCP server integration

jubnl:
Adds a Model Context Protocol (MCP) server to TREK, letting users connect AI assistants (Claude, Cursor, etc.) directly to the trip data and control the app through natural language. 

The endpoint lives at `/mcp` and supports the full Streamable HTTP transport. Authentication uses long-lived API tokens (`trek_*` prefix) generated from the Settings page. All the details on available resources, tools, setup instructions, and limitations are documented in [MCP.md](./MCP.md).

Tools and resources:
- **14 read-only resources**: browse trips, places, budget, packing lists, reservations, day notes, collab notes, bucket list, visited countries, and more
-  **34 tools**: create/update/delete trips, places, day assignments, reservations, budget items, packing items, notes, and atlas entries
-  `get_trip_summary` loads everything about a trip in one shot, which is useful as a context primer before asking the assistant to do anything

Changes are broadcast to WebSocket clients in real time, so edits made through MCP show up immediately in the browser. 

Regarding security:
- Tokens are stored as SHA256 hashes, the raw value is shown once at creation time
- 60 requests/minute per user, max 5 concurrent sessions, 1-hour inactivity TTL
- Max 10 tokens per user
- All writes are blocked in demo mode
- The MCP addon is disabled by default, admins enable it per instance
- Tokens are linked to a user and respect the user's rights

Changes:
- `server/src/mcp/`: session handler, resources, and tools
- `server/src/routes/auth.ts`: token CRUD endpoints (`GET/POST/DELETE /api/auth/mcp-tokens`)
- `server/src/routes/admin.ts`: admin token management
- `server/src/db/migrations.ts`: migrations for  `mcp_tokens` table + addon entry
- `client/src/pages/SettingsPage.tsx`: token management UI with copy-to-clipboard config snippet
- `client/src/components/Admin/AdminMcpTokensPanel.tsx`: admin panel
- Translation keys added across all 8 languages

I think I tested every possible case (at least that I could think of), but please review the code carefully as it is quite an addition to TREK.

### DIFF
--- a/MCP.md
+++ b/MCP.md
@@ -12,7 +12,7 @@ structured API.
 - [Limitations & Important Notes](#limitations--important-notes)
 - [Resources (read-only)](#resources-read-only)
 - [Tools (read-write)](#tools-read-write)
-- [Screenshots](#screenshots)
+- [Example](#example)
 
 ---
 


kbzowski:
I wonder what the advantages of this approach are, given that TREK has a REST API that supports JWT authentication?

jubnl:
> I wonder what the advantages of this approach are, given that TREK has a REST API that supports JWT authentication?

I'm not 100% sure if you mean the implementation of the MCP or the MCP itself. Let's answer to both:

Regarding the implementation:
I implemented the MCP this way because I didn't wanted to derive the tokens to a JWT token cause it introduce a "hop". If that was the case it means that first the JWT token should be generated, given back to the LLM and then either make a proxy from the MCP to the REST API or the LLM directly call the API. Either way it means that you give an LLM a JWT token. That's a huge security concerns IMO a JWT token is a user token, meaning all the endpoints that a real user can call, the LLM will be able to call. Even private settings, access to the full user profile unrestricted and so on.

Regarding the MCP itself:
Yeah I don't like that too. People needed a standardized way to use tools in LLM and MCPs were what people came up with. I don't personally like it and I would prefer a standard web API (rest, soap, graphql or whatever) but MCP is now a standard, which I'm not really fund of TBH.

Let me know if that answers your question

swheettaos:
This is me asking a legit question here .. trying to learn something.  I notice in the resources and tools you are accessing the DB directly.  In my first attempt at this, I started the same way, but I thought .. do I like that it's touching the DB directly .. why have it query/modify the DB directly (that seemed dangerous).. why not have it's use the same API's the that the client is using.  That way if the backend changes the API may account for that and have better chance of not having something break for the the MCP.. in my head it also seems like managing code for things that touch the DB directly in multiple places is asking for trouble.  Using the the API would result in a worse case it just breaks the MCP if the API changes rather than the MCP changing the DB in ways it shouldn't be anymore.  At least this sounded better in my head.  

If I'm mis-reading your code .. please let me know also .. like I said.. trying to learn here.

It does mean there is an extra internal hop, but I thought it would be worth it. To ensure the MCP is using the existing API calls instead of directly modifying the DB .. This is me trying to learn .. not trying to rip on your stuff.

kbzowski:
My Trek-MCP attempt was based on direct API access (https://github.com/kbzowski/trek-mcp-server). JWT should be short-lived enough to avoid security issues. Of course, direct access to the entire API as admin is dangerous, but this is a calculated risk when using LLM today. It's good that Trek supports automatic backups.

swheettaos:
@kbzowski I think the approach that @jubnl is doing with per user tokens (if I reading it correctly) is better than the entire api admiin .. it allows each users to use things like claude or something else to interact but restricts them to only the things they should have access too.. This is the approach I was trying to take in my feable attempt also.  Mine had other issues (I'm learning).  The sessions and token handling is pretty robust in @jubnl .. The only difference in architecture in mine was I was trying to leverage the existing APIs rather than talking to the DB directly to get and modify.  This was the question I was asking @jubnl.  I might be reading his code wrong though.

mauriceboe:
Merged — great work @jubnl, this is an excellent contribution. Clean architecture, solid security model, and well-documented.

I resolved the merge conflicts with v2.7.1 and added the missing MCP translation keys for the 3 languages that were added after your PR was branched (Czech, Hungarian, Italian).

---

For context on why we went with this built-in approach over the external [trek-mcp-server](https://github.com/kbzowski/trek-mcp-server) by @kbzowski:

**Security:** This PR uses hashed long-lived API tokens (shown once, stored as SHA256) with per-user scoping and rate limiting. The external package requires storing email + password in plaintext environment variables and authenticates via JWT login — no MFA support, no token revocation, no session limits.

**Integration:** This runs inside the TREK server process, which means changes made through MCP are broadcast to all connected WebSocket clients in real-time. The external package is a separate process that calls the HTTP API — no live sync awareness, no addon gate, no admin control.

**Admin control:** This is an addon that's disabled by default. Admins can enable/disable it, view all tokens across users, and revoke access instantly. The external package has no visibility or control from the TREK admin panel.

**User experience:** Token management UI in Settings with copy-paste client config snippets. The external package requires manual environment variable setup.

That said, the external package is a solid standalone option for users who want MCP without modifying their TREK instance — both approaches have their place.

jubnl:
The biggest factor that made me chose this architecture was that I could make a hard restriction on the data that an LLM can get/post/put/delete.
There's other ways that could be achieved but it would require a pretty big refactor of TREK with a strict separation between services and controllers (routes). If the app had such separation, I could just reuse the services (which are technically internal APIs).

Here's an example of such separation:

route:
```ts
router.get('/users', (req: Request, res: Response) => {
  const usersWithStatus = userService.getUsersWithStatus();
  res.json({ users: usersWithStatus });
});
```

service:
```ts
export function getUsersWithStatus(): any { // any just for the example but it should return structured data
  const users = db.prepare(
    'SELECT id, username, email, role, created_at, updated_at, last_login FROM users ORDER BY created_at DESC'
  ).all() as Pick<User, 'id' | 'username' | 'email' | 'role' | 'created_at' | 'updated_at' | 'last_login'>[];
  let onlineUserIds = new Set<number>();
  try {
    const { getOnlineUserIds } = require('../websocket');
    onlineUserIds = getOnlineUserIds();
  } catch { /* */ }
  const usersWithStatus = users.map(u => ({
    ...u,
    created_at: utcSuffix(u.created_at),
    updated_at: utcSuffix(u.updated_at as string),
    last_login: utcSuffix(u.last_login),
    online: onlineUserIds.has(u.id),
  }));
  return usersWithStatus;
}
```

If the entire app was made like this, I could just reuse the services directly without making any db calls in the MCP. But it require a (almost) complete rewrite of the backend.

Since the controllers are also services, it means that the methods are locked behind auth (JWT). A JWT token give access to the entire user account, which I'm not comfortable at all to give to a LLM.

I hope that explanation helps @swheettaos !

swheettaos:
@jubnl yes that helps .. so while I had the right idea.. i wasn't implmenting it right.  I thought by calling the api with JWT token from the frontend (or I thought thats how the call would be made) that the it would only give them the same access that they get when using the frontend.  I missed the boat on that i guess.   I had other issues in my code too but I should've have made it more clear that my code wasn't finished nor tested .. it was supposed to be more about "what do you think about this approach".   Utimately what you put together was what I was trying to do with almost the same methodolgy (internal, per user, admin enable/disable, admin controle over tokens, restricted access based on app permissions .. just poor implementation.

jubnl:
@swheettaos You'll know for next time but basically, open pull request means that a patch was made for the software and it is ready for code review and potentially a merge. Basically prod ready code which should by properly reviewed and tested by the one who created the patch in the 1st place. There's a ton of repositories that just closed their PR entirely as it was spammed by people that just created half baked pr with no code review and no testing at all. They lost too much time reviewing everything. Take a look at the openclaw repo if you want an example of pr spam.

swheettaos:
Fair point — I should have opened a feature request and started the conversation there. I'm used to either working solo on my code or in teams with direct communication with structured tools like Jira/Kanban. Good reminder to follow the process.

Never my intention to just throw crap over the wall.

M-Enderle:
Added small improvements in #191

PR #130, opened 2026-03-30
feat: Add support for Czech language

Numira-code:
adds Czech (cs) as another language. complete translations for all keys


mauriceboe:
Merged to `dev`, added missing keys to match en.ts (~479 keys synced). Thanks @Numira-code! Shipping today.

PR #134, opened 2026-03-30
fix: use i18n key for admin page title

slashwarm:
the admin page h1 is hardcoded as "Administration" instead of using the existing t('admin.title') key. one line fix.


mauriceboe:
Hey @slashwarm! Good catch — we already fixed this independently on `dev`. Same one-liner. Thanks for spotting it though, shipping today!

PR #138, opened 2026-03-30
fix: filename encoding and type annotation for linksMap

asteriskyg:
### Non-ASCII filenames corrupted on upload

**Affected files:** `server/routes/files.ts`, `server/routes/collab.ts`

Files uploaded with non-ASCII filenames (Korean, Japanese, etc.) had their `original_name` saved as garbled characters in the database.

multer reads the `filename` parameter from the `Content-Disposition` header as Latin-1, while browsers send it as raw UTF-8 bytes — a known issue ([#1104](https://github.com/expressjs/multer/issues/1104)) fixed in v2.1.0 ([#1210](https://github.com/expressjs/multer/pull/1210)).

Since the option defaults to `'latin1'` for backward compatibility, it must be explicitly set.
Adding `defParamCharset: 'utf8'` to the multer config resolves this at the source.

**Before:**
```
original_name: "áá¡á¼áá©á¼áá¯á«.pdf"
```
**After:**
```
original_name: "항공권.pdf"
```

---

### Incorrect `linksMap` type annotation

`linksMap` was typed as `Record<number, number[]>` but actually stored full link objects, causing a TypeScript error.

Fixing the type to use a `FileLink` interface resolves this.

### DIFF
--- a/server/src/routes/files.ts
+++ b/server/src/routes/files.ts
@@ -73,6 +73,12 @@ function formatFile(file: TripFile) {
 }
 
 // List files (excludes soft-deleted by default)
+interface FileLink {
+  file_id: number;
+  reservation_id: number | null;
+  place_id: number | null;
+}
+
 router.get('/', authenticate, (req: Request, res: Response) => {
   const authReq = req as AuthRequest;
   const { tripId } = req.params;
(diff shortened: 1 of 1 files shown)


mauriceboe:
Hey, clean fix. The defParamCharset option is the correct solution for the well-known multer Latin-1 encoding issue, and the FileLink type annotation matches what the code actually stores. Merged into dev.

Issue #148, opened 2026-03-30
[BUG] OIDC login causes redirect loop in oidc-only mode


Xre0uS:
**Describe the bug**
When `OIDC_ONLY=true` is set, logging in via OIDC results in an infinite redirect loop and the user is never authenticated.

**To Reproduce**
1. Configure TREK with a valid OIDC provider (Authelia in my case) and `OIDC_ONLY=true`
2. Navigate to the login page and gets redirected to the OIDC provider
3. Complete authentication at the provider
4. Provider redirects back to `/login?oidc_code=<code>`
5. The page immediately redirects back to `/api/auth/oidc/login` and loop begins

**Expected behavior**
After completing OIDC authentication, the user should be able to access the Trek dashboard.

**Desktop (please complete the following information):**
 - OS: Windows, Ubuntu
 - Browser Firefox
 - Version 149.0

mauriceboe:
Fixed via #149 — merged to `dev`, shipping today. Thanks @Xre0uS!

Jeppedy:
> **Describe the bug** When `OIDC_ONLY=true` is set, logging in via OIDC results in an infinite redirect loop and the user is never authenticated.
> 
> **To Reproduce** Steps to reproduce the behavior:
> 
> 1. Configure TREK with a valid OIDC provider (Authelia in my case) and `OIDC_ONLY=true`
> 2. Navigate to the login page and gets redirected to the OIDC provider
> 3. Complete authentication at the provider
> 4. Provider redirects back to `/login?oidc_code=<code>`
> 5. The page immediately redirects back to `/api/auth/oidc/login` and loop begins
> 

What is the proper Redirect URI to use in Authentik?

Xre0uS:
@Jeppedy 

Issuer URL: `https://your-authentik.com/application/o/trek/`
Callback URL: `https://your-domain.com/api/auth/oidc/callback`

https://github.com/mauriceboe/TREK/wiki/Single-Sign-On#authentik

PR #149, opened 2026-03-30
fix: prevent OIDC redirect loop in oidc-only mode

Xre0uS:
Previously `getAppConfig()` was fired at the top of the effect and resolved after `window.history.replaceState` had already cleared `?oidc_code=` from the URL, causing the `oidc_only_mode` redirect guard to pass incorrectly. Moving the config fetch below the early-return blocks makes it unreachable during any callback flow.

Changed `LoginPage.tsx` , reordered the `useEffect` logic so callback query params (`oidc_code`, `oidc_error`, `invite`) are handled first with early `return`s, and the `getAppConfig()` auto-redirect only runs when none of those params are present.

Additional cleanup in the same pass:
- `navigate('/dashboard', { replace: true })` instead of `navigate` + `window.location.reload()`
- `setIsLoading(true/false)` wrapping the OIDC token exchange fetch
- Fixed `useEffect` dependency array (`navigate`, `t`)

Fixes https://github.com/mauriceboe/TREK/issues/148

### DIFF
--- a/client/src/pages/LoginPage.tsx
+++ b/client/src/pages/LoginPage.tsx
@@ -33,25 +33,12 @@ export default function LoginPage(): React.ReactElement {
   const navigate = useNavigate()
 
   useEffect(() => {
-    authApi.getAppConfig?.().catch(() => null).then((config: AppConfig | null) => {
-      if (config) {
-        setAppConfig(config)
-        if (!config.has_users) setMode('register')
-        // Auto-redirect to OIDC if password auth is disabled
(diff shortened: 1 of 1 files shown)


mauriceboe:
Clean fix, merged to `dev`. Thanks @Xre0uS — the reordering of the useEffect logic with early returns is exactly the right approach. The race condition between `getAppConfig()` and the OIDC code exchange was subtle. Appreciate the contribution!

PR #158, opened 2026-03-30
added italian translation

entropyst72:


mauriceboe:
Merged to `dev`, synced with all other keys. Thanks @entropyst72! Shipping today.

Issue #155, opened 2026-03-30
[FEATURE] Require users to set 2FA auth during registration


ScDor:
**Is your feature request related to a problem? Please describe.**
When not using an OIDC, users may set a password-only auth, without 2FA.

**Describe the solution you'd like**
Admins should be able to require users to add 2FA.

**Describe alternatives you've considered**
- using OIDC
- asking users manually

mauriceboe:
Hey @ScDor! Good security suggestion.

@fgbona you built the 2FA system — would you be interested in adding an admin toggle to enforce 2FA for all users? The idea would be:

1. Admin setting "Require 2FA" in Admin > Settings
2. Users without 2FA configured get redirected to the 2FA setup page after login
3. Can't access the app until 2FA is set up

If you're up for a PR on this, that would be great!

fgbona:
Hi @mauriceboe I will work on this request.

mauriceboe:
Implemented via PR #166 and PR #169 by @fgbona:

- **Admin toggle** in Admin → Settings to require 2FA for all users
- Users without MFA are redirected to Settings and blocked from the app until setup is complete
- Server-side enforcement via middleware (API + WebSocket)
- **MFA backup codes** — 10 one-time recovery codes generated on enable, with copy/download/print options
- Backup codes accepted at login alongside TOTP codes
- Admin must have MFA enabled before enforcing the policy
- Users cannot disable MFA while the policy is active

Available in the latest version.

PR #166, opened 2026-03-30
feat(require-mfa): #155 enforce MFA via admin policy toggle across app access

fgbona:
Add an admin-controlled `require_mfa` policy in App Settings and expose it via `/auth/app-config` so the client can enforce it globally. Users without MFA are redirected to Settings after login and blocked from protected API/WebSocket access until setup is completed, while preserving MFA setup endpoints and admin recovery paths. Also prevent enabling the policy unless the acting admin already has MFA enabled, and block MFA disable while the policy is active. Includes UI toggle in Admin > Settings, required-policy notice in Settings, client-side 403 `MFA_REQUIRED` handling, and i18n updates for all supported locales.


mauriceboe:
Merged — thanks @fgbona, another solid contribution\! Clean middleware design with well-thought-out exemptions (MFA setup endpoints, OIDC flow, demo mode). The guard rails are great too: admin must have MFA enabled before enforcing, and users can't disable MFA while the policy is active. Appreciate you covering all 12 languages as well.

PR #169, opened 2026-03-30
fix(mfa-backup-codes): persist backup codes panel after enable and refresh

fgbona:
Keep MFA backup codes visible after enabling MFA by avoiding protected-route unmount during user reload (`loadUser({ silent: true })`) and restoring pending backup codes from sessionStorage until the user explicitly dismisses them.


mauriceboe:
Merged — thanks again @fgbona! MFA backup codes are a critical feature for any 2FA implementation. Nice touches with the sessionStorage persistence (so codes survive the loadUser reload), the silent loadUser option to avoid unmounting protected routes, and the copy/download/print actions. Also fixed a small syntax issue in the migration (missing closing bracket) during conflict resolution.

PR #175, opened 2026-03-30
feat: add per-item currency support with server-side exchange rate co…

### LINKED ISSUE #38: [FEATURE] Currency convert when switched in app
**Is your feature request related to a problem? Please describe.**
I'm always frustrated when I switch currencies and it doesn't convert from the displayed to the new currency selected.

**Describe the solution you'd like**
Set a base currency for a trip. If different currencies are selected after that it should convert to make it easier to plan and see prices.

**Describe alternatives you've considered**
I've considered putting everything in one currency but unfortunately the place I'm traveling to uses a different currency. i.e. I am traveling from Japan to the Philippines and trying to figure out food prices and local stays but they are both in local currency so I have to convert those into my local currency to place it in the app.

**Additional context**

BKSalman:
resolves #38

these changes were AI assisted, since I'm currently on a trip :sweat_smile:, but I would really appreciate this feature being added

if there is anything you don't like or think is AI slop, I can manually change it

### DIFF
--- a/client/src/components/Planner/ReservationModal.tsx
+++ b/client/src/components/Planner/ReservationModal.tsx
@@ -11,6 +11,7 @@ import { useTranslation } from '../../i18n'
 import { CustomDatePicker } from '../shared/CustomDateTimePicker'
 import CustomTimePicker from '../shared/CustomTimePicker'
 import type { Day, Place, Reservation, TripFile, AssignmentsMap, Accommodation } from '../../types'
+import { CURRENCIES } from '../../utils/formatters'
 
 const TYPE_OPTIONS = [
   { value: 'flight',     labelKey: 'reservations.type.flight',     Icon: Plane },
@@ -74,6 +75,7 @@ export function ReservationModal({ isOpen, onClose, onSave, reservation, days, p
 
   const isBudgetEnabled = useAddonStore(s => s.isEnabled('budget'))
   const budgetItems = useTripStore(s => s.budgetItems)
+  const tripCurrency = useTripStore(s => s.trip?.currency || 'EUR')
   const budgetCategories = useMemo(() => {
     const cats = new Set<string>()
     budgetItems.forEach(i => { if (i.category) cats.add(i.category) })
@@ -84,13 +86,12 @@ export function ReservationModal({ isOpen, onClose, onSave, reservation, days, p
     title: '', type: 'other', status: 'pending',
     reservation_time: '', reservation_end_time: '', end_date: '', location: '', confirmation_number: '',
     notes: '', assignment_id: '', accommodation_id: '',
-    price: '', budget_category: '',
-    meta_airline: '', meta_flight_number: '', meta_departure_airport: '', meta_arrival_airport: '',
-    meta_departure_timezone: '', meta_arrival_timezone: '',
-    meta_train_number: '', meta_platform: '', meta_seat: '',
-    meta_check_in_time: '', meta_check_out_time: '',
-    hotel_place_id: '', hotel_start_day: '', hotel_end_day: '',
-  })
+        price: '', budget_category: '', budget_currency: '',
+        meta_airline: '', meta_flight_number: '', meta_departure_airport: '', meta_arrival_airport: '',
+        meta_departure_timezone: '', meta_arrival_timezone: '',
+        meta_train_number: '', meta_platform: '', meta_seat: '',
+        meta_check_in_time: '', meta_check_out_time: '',
+      })
   const [isSaving, setIsSaving] = useState(false)
   const [uploadingFile, setUploadingFile] = useState(false)
   const [pendingFiles, setPendingFiles] = useState([])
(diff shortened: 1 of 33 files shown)


mauriceboe:
Hey, nice work on this. The overall design is solid: per-item currency with server-side conversion, SQLite-cached rates with stale fallback, WebSocket propagation, and all aggregations correctly using COALESCE(converted_price, total_price).

Found one bug though: `recalculateTrip` in `exchangeRates.ts` fetches rates for the base currency (e.g. EUR) via `fetchAndCacheRates(baseCurrency)`, which caches EUR→JPY, EUR→USD etc. But then it looks up `getRate(itemCur, baseCurrency)` which searches for e.g. JPY→EUR — that direction isn't in the cache, so it returns null. Individual item creation works fine because `convertAmount` correctly calls `fetchAndCacheRates(from)`, but `recalculateTrip` (called on currency change and manual refresh) would set `converted_price = null` for all non-base-currency items.

Simplest fix: use `convertAmount(item.total_price, itemCur, baseCurrency)` per item inside `recalculateTrip` instead of the manual rate lookup, or fetch rates for each unique item currency rather than just the base currency.

Also the package-lock.json changes add `"peer": true` to a bunch of unrelated packages, probably from a different npm version. Would be good to clean that up to keep the diff focused.

BKSalman:
ok, now it should be fixed, with other small bug fixes, and I also added the tables to the schema.ts file

please tell me if there is anything else wrong with it

mauriceboe:
Hey, thanks for the update – the `recalculateTrip` bug is fixed correctly now, nice work.

Found a couple more things though:

1. **Schema drift with `paid_by_user_id`**: It's in `schema.ts` (CREATE TABLE) but not in the migration. Fresh databases will have the column, existing ones won't. Since it's unused right now, I'd just remove it from `schema.ts` until it's actually needed – otherwise it'll cause confusion down the line.

2. **No validation on `item_currency`**: The value from `req.body` goes straight into the DB and the API URL without any check. At minimum validate it's a 3-letter uppercase string, ideally check against a currency list.

3. **No timeout on the external fetch**: If exchangerate-api.com hangs, the request hangs forever. An `AbortSignal.timeout(10000)` or similar would be good.

These are pretty quick fixes. Once 1 and 2 are sorted I'm happy to merge.

BKSalman:
@mauriceboe Thank you for your time, it should be good now

BKSalman:
just resolved conflicts, and tested the basic functionality, and it seems like it's working

mauriceboe (changes requested):
Hey, nice work on this. The overall approach is solid and the earlier feedback was addressed well. Found a few more things though:

### Bugs

**1. `item_currency` not normalized on update**
`createBudgetItem` correctly does `.toUpperCase()`, but `updateBudgetItem` writes `data.item_currency` straight into the DB without normalizing. If someone sends `"eur"` via PUT, it lands lowercase in the DB. The route validation (`isValidCurrency`) accepts it because it checks `val.toUpperCase()` but passes the original value through. Easy fix: normalize in `updateBudgetItem` before passing to the query.

**2. Race condition in `recalculateTrip` – parallel API fetches**
`Promise.all` over all items means if 10 items share the same foreign currency, all 10 hit the cache miss simultaneously and fire 10 parallel `fetchAndCacheRates("JPY")` calls before the first one finishes caching. Could trigger rate limits on the external API.

Fix: collect unique item currencies first, fetch them sequentially (or deduplicated), then convert locally.

### Security

**3. `refresh-rates` doesn't check `budget_edit` permission**
The endpoint only checks `verifyTripAccess` – any trip member (even read-only) can spam rate refreshes and hammer the external API. Should either require `budget_edit` or have some rate limiting.

### Cleanup

- `formatConvertedAmount` in `formatters.ts` is exported but never imported anywhere – dead code.
- The `package-lock.json` still has the unrelated `"peer": true` additions from the earlier review.

1–3 should be addressed before merge, the rest is nice-to-have.

BKSalman:
@mauriceboe ok, done :+1:

BKSalman:
fixed failing tests

BKSalman:
resolved conflicts, and removed `"peer": true` from `package-lock.json`

server/package-lock.json
jubnl, line 1: Please remove all the "peer": true

server/src/db/migrations.ts
jubnl, line 1: three existing comments were silently removed from migrations.ts (lines for Todo items, per-day positions, etc.). These are unrelated deletions by the AI. Should be restored.

server/src/routes/budget.ts
```diff
@@ -135,6 +146,33 @@ router.get('/settlement', authenticate, (req: Request, res: Response) => {
   res.json(calculateSettlement(tripId));
 });
 
+router.post('/refresh-rates', authenticate, async (req: Request, res: Response) => {
```
jubnl, line 149: a user can spam this, triggering repeated external API calls. A 60-second cooldown per trip would suffice

client/src/store/slices/remoteEventHandler.ts
```diff
@@ -214,6 +215,11 @@ export function handleRemoteEvent(set: SetState, event: WebSocketEvent): void {
               : i
           ),
         }
+      case 'budget:rates-updated': {
```
jubnl, line 218: the budget:rates-updated handler calls get().loadBudgetItems() (an API call from inside the reducer), which is architecturally inconsistent with every other handler in the file. The WebSocket event could carry the updated items directly

client/src/components/Planner/ReservationModal.tsx
```diff
@@ -84,13 +86,12 @@ export function ReservationModal({ isOpen, onClose, onSave, reservation, days, p
...
     reservation_time: '', reservation_end_time: '', end_date: '', location: '', confirmation_number: '',
     notes: '', assignment_id: '', accommodation_id: '',
-    price: '', budget_category: '',
-    meta_airline: '', meta_flight_number: '', meta_departure_airport: '', meta_arrival_airport: '',
-    meta_departure_timezone: '', meta_arrival_timezone: '',
-    meta_train_number: '', meta_platform: '', meta_seat: '',
-    meta_check_in_time: '', meta_check_out_time: '',
-    hotel_place_id: '', hotel_start_day: '', hotel_end_day: '',
-  })
+        price: '', budget_category: '', budget_currency: '',
```
jubnl, line 89: mixed indentation in the form state

jubnl (changes requested):
Other things to address:
- Clean up unused translation keys
- Add a comment in exchangeRates.ts noting the free tier limits of api.exchangerate-api.com/v4 (~1500 req/month) so future maintainers have context.

BKSalman:
I will close this for now, and open a new PR when I have time to do this properly, thank you for your time

If anyone is interested in doing this before I do, please feel free to do so

PR #177, opened 2026-03-31
Add flight API

Stew8260:
Add Flight API integration using AeroDataBox

This PR adds flight information functionality to the TREK project by integrating the AeroDataBox API via RapidAPI. It enables fetching flight data for flights, including departures and arrivals.

Integration with AeroDataBox API (RapidAPI)
Ability to fetch flight data by airport (ICAO code)
Support for departure/arrival time windows (like departing on day 1 and arriving on day 2)
Basic flight data retrieval (flight number, destination, time, airline

### DIFF
--- a/client/src/components/Planner/DayPlanSidebar.tsx
+++ b/client/src/components/Planner/DayPlanSidebar.tsx
@@ -639,6 +639,7 @@ export default function DayPlanSidebar({
 
   const handleDropOnDay = (e, dayId) => {
     e.preventDefault()
+    e.stopPropagation()
     setDragOverDayId(null)
     const { placeId, assignmentId, noteId, fromDayId } = getDragData(e)
     if (placeId) {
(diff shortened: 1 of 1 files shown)


jubnl:
watch out, you committed your `client/pnpm-lock.yaml` which isnt part of the original project as it uses npm as its package manager

Stew8260:
Fixed

mauriceboe:
Hey, the flight search feature itself is a good idea and the UX flow (search first, manual fallback) is well thought out. The arc visualization on the map is a nice touch too.

However the PR needs some cleanup before it can be merged:

There's a `TEST123` left in README.md.

A large chunk of the diff is just reformatting/re-indentation of existing code (DayPlanSidebar alone is 328+/310- lines of mostly whitespace changes). This makes it very hard to review the actual feature changes. Please separate formatting changes or revert them so the diff only shows the real changes.

The server uses `node-fetch` everywhere else, but this PR adds `axios` as a new dependency for the flight service. Please stick with `node-fetch` to keep things consistent.

Several TypeScript types are loosened from specific types to `Record<string, any>` and `Promise<any>`. This should go the other direction if anything.

The CSS import changes from `leaflet.markercluster` to `react-leaflet-cluster` paths with `@ts-ignore` are unrelated to flights and could break existing styling. Should be a separate PR if needed.

The RapidAPI key validation hits `/subscriptions/webhook` and just checks for non-401/403 status, which is fragile. A dedicated health check endpoint would be more reliable.

Would love to see a cleaned up version of this.

jubnl:
btw @henkie2543545 you should definitely change all of your passwords since you installed axios on your machine

https://snyk.io/blog/axios-npm-package-compromised-supply-chain-attack-delivers-cross-platform/

PR #180, opened 2026-03-31
fix: weather API reliability - cache, request queue, and IPv4 agent

shanelord01:
- Add file-based JSON cache for weather API responses (permanent for historical data, 6hr TTL for forecasts)
- Replace concurrent Promise.all weather fetches with sequential queue (one request at a time, 300ms delay between requests)
- Consolidate overlapping date ranges for the same coordinates into single API requests
- Add retry with exponential backoff for transient network errors
- Force IPv4 via custom HTTPS agent to prevent ETIMEDOUT on Docker bridge networks where IPv6 is unroutable
- Add NODE_OPTIONS=--dns-result-order=ipv4first to docker-compose example as defence in depth

Claude Code used for this. Please disregard, use, rewrite or ignore - no offence intended in any way!

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -26,6 +26,7 @@ services:
       # - ALLOWED_ORIGINS=https://yourdomain.com  # Optional: restrict CORS to specific origins
       - PORT=3000
       - TZ=${TZ:-UTC}
+      - NODE_OPTIONS=--dns-result-order=ipv4first
     volumes:
       - ./data:/app/data
       - ./uploads:/app/uploads
--- a/server/src/routes/weather.ts
+++ b/server/src/routes/weather.ts
@@ -1,6 +1,6 @@
 import express, { Request, Response } from 'express';
-import fetch from 'node-fetch';
 import { authenticate } from '../middleware/auth';
+import { weatherFetch } from '../services/weatherFetch';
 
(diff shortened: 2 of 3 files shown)


mauriceboe:
Hey, appreciate the effort here. The core problem (ETIMEDOUT in Docker) is real, but the solution is heavily over-engineered for what's needed.

The IPv4 agent and dns-result-order fix alone would solve the actual ETIMEDOUT issue in about 5 lines. The remaining 520 lines add a file-based cache, sequential request queue, archive request consolidation with coordinate rounding and response slicing, and retry with backoff. That's a lot of custom infrastructure for a problem that likely doesn't exist once the IPv4 fix is in place.

Some specific concerns with the current approach: the sequential queue with 300ms delay between all requests adds significant artificial latency (a trip with 10 days = 3+ seconds of waiting). The file cache for historical data never expires and has no size limit, so it grows forever. The request consolidation logic (50ms buffer, coordinate grouping, date range merging, response slicing) is complex but would almost never trigger in practice since multiple archive requests for the same coordinates rarely arrive within 50ms.

Would be great to see a smaller PR that just adds the IPv4 agent fix and the docker-compose change. If rate limiting turns out to be an issue after that, a simple in-memory cache with TTL would be the next step.

PR #181, opened 2026-03-31
fix: update dayAccommodations state after create/edit/delete

BKSalman:
when adding an accommodation by clicking the "Add accommodation" button in the day details panel, it wasn't updating the state in the UI, this resolves that issue


mauriceboe:
Good catch. The dayAccommodations state wasn't being updated after mutations, so the UI didn't reflect changes until re-opening the day. Clean fix. Merged into dev.

PR #182, opened 2026-03-31
mobile UI fixes

BKSalman:
this PR fixes the following:
- day details on mobile not showing
- the add note modal z-index

and also makes some of the icons bigger in the day planing panel, and in the add note modal 
 before (very hard to press especially on mobile): 
image

after (much easier to press):
image


mauriceboe:
Nice work, all three fixes are clean and well targeted. The matchMedia approach for mobile detection is the right call. Merged into dev, thanks!

PR #185, opened 2026-03-31
fix(server): encrypt api keys

### LINKED ISSUE #167: [ISSUE] Google Map API key
Issue:
Currently, the Google Maps API key is stored in plain text. 

Proposed solution:
Store the API key encrypted

Akashic101:
This fixes #167 by encrypting the api keys for Google Places and OpenWeather in the db, making them unreadable unless the proper JWT-Token of the client is provided

### DIFF
--- a/server/src/routes/auth.ts
+++ b/server/src/routes/auth.ts
@@ -17,6 +17,7 @@ import { randomBytes, createHash } from 'crypto';
 import { revokeUserSessions } from '../mcp';
 import { AuthRequest, User } from '../types';
 import { writeAudit, getClientIp } from '../services/auditLog';
+import { decrypt_api_key, maybe_encrypt_api_key } from '../services/apiKeyCrypto';
 
 authenticator.options = { window: 1 };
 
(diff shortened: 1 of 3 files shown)


Akashic101:
Note that this does not change anything about previously stored unencrypted keys. Users will need to copy the keys, save it so it gets removed, then paste it in again to encrypt the keys. Also if the JWT-Token would get leaked the api-key could be stolen as there is no rotation at the moment, however this is pretty unlikely and implement this would break the scope of this PR

mauriceboe:
Hey, took a thorough look at this PR. The crypto implementation is solid: AES-256-GCM with random IV, auth tag, clean key derivation with domain separator, and a versioned prefix for future algorithm changes. The integration into the existing endpoints is clean, and legacy plaintext keys are correctly detected and continue to work.

Two things I noticed:

An automatic migration on server start would be nice so existing plaintext keys don't have to wait for the user to manually re-save them. A simple pass over all rows at startup that encrypts unencrypted keys would solve this.

The extended maps_details error diagnostics in the validate-keys endpoint is unrelated to encryption and would be better suited for a separate PR.

Merged the branch into dev for testing.

mauriceboe:
Merged into dev manually, closing.

PR #189, opened 2026-03-31
feat(add-gpx-tracks): adds better gpx track views

M-Enderle:
Lets users import routes like hikes into the application with basic information read directly from the gpx file such as distance, min elevation, max elevation. Also displays a detail elevation graph

image

(previous pull accidentally closed)

### DIFF
--- a/client/package-lock.json
+++ b/client/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "trek-client",
-  "version": "2.7.0",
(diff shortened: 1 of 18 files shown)


mauriceboe:
Great improvement over the old start/end marker approach. Full route geometry on the map plus the elevation profile and track stats make GPX imports actually useful now. Merged into dev, thanks!

PR #195, opened 2026-03-31
feat(atlas): add searchbar

Akashic101:
This PR adds a searchbar in the atlas to allow the easier location of countries. This also fixes the issue of small countries, Vatican City as an example, not beeing selectable on the map

image
image


Akashic101:
The translations may not be perfect for some countries and have been automatically translated, also unsure why it shows such a big diff for `cs` and `it`

mauriceboe:
Nice feature, the search bar solves both discoverability and the small-country selection problem cleanly. Merged into dev without the cs.ts/it.ts line-ending changes to keep the diff clean. Thanks!

mauriceboe:
Credit: feature by @Akashic101

PR #238, opened 2026-03-31
feat: configurable permissions system in admin

slashwarm:
## Summary

- **Admin panel** — new Permissions tab lets admins configure who can perform each action (create/edit/delete trips, upload files, manage members, edit places, budget, packing, reservations, collab, share links) across four levels: admin-only, trip owner, trip members, or everyone
- **Server enforcement** — `checkPermission()` applied to all mutating routes; settings stored in `app_settings` with `perm_*` prefix; in-memory cache invalidated on save; backwards-compatible defaults match upstream behaviour
- **Client gating** — `useCanDo()` Zustand hook reads permissions loaded at startup; no prop drilling; gates UI in DashboardPage, FileManager, PlacesSidebar, TripMembersModal
- **i18n** — 46 `perm.*` translation keys added to all 12 language files

## Technical notes

- `TripOwnerContext` type in `permissionsStore` accepts both `Trip` (has `owner_id`) and `DashboardTrip` (has `is_owner` + index signature) without casting at call sites
- Redundant DB queries eliminated: `canAccessTrip()` result reused for owner/member context in `trips.ts` and `files.ts` instead of separate `getTripOwnerId()` + `isOwner()` calls
- `requireTripAccess` middleware result (`authReq.trip`) reused in file upload route — no second access check

### DIFF
--- a/client/src/components/Collab/CollabNotes.tsx
+++ b/client/src/components/Collab/CollabNotes.tsx
@@ -222,13 +222,15 @@ interface NoteFormModalProps {
   existingCategories: string[]
   categoryColors: Record<string, string>
   getCategoryColor: (category: string) => string
-  note?: CollabNote | null
-  tripId?: number
+  note: CollabNote | null
+  tripId: number
   t: (key: string) => string
-  canUploadFiles?: boolean
 }
 
-function NoteFormModal({ onClose, onSubmit, onDeleteFile, existingCategories, categoryColors, getCategoryColor, note, tripId, t, canUploadFiles = true }: NoteFormModalProps) {
+function NoteFormModal({ onClose, onSubmit, onDeleteFile, existingCategories, categoryColors, getCategoryColor, note, tripId, t }: NoteFormModalProps) {
+  const can = useCanDo()
+  const tripObj = useTripStore((s) => s.trip)
+  const canUploadFiles = can('file_upload', tripObj)
   const isEdit = !!note
   const allCategories = [...new Set([...existingCategories, ...Object.keys(categoryColors || {})])].filter(Boolean)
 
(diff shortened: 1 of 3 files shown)


slashwarm:
I found some issues, a few minutes please

slashwarm:
Thank you, ready for review

mauriceboe:
Hey, thanks for the work – looks solid overall. A few things before I merge:

**Must be clarified:**

1. Are the `GET/PUT /permissions` routes in `admin.ts` protected by `adminOnly` through the parent router? I don't see it explicitly in the diff – if not, any authenticated user could modify the global permissions.

2. The defaults for `trip_edit` and `member_manage` are set to `trip_member` – previously only the owner/admin could do that. Was this intentional? I'd feel safer with `trip_owner` as the default, anyone who wants it more open can change it in the panel.

**Should be fixed:**

3. In `DayPlanSidebar`, the context menu becomes empty when `canEditDays` is false – either suppress the menu entirely or add a "View" option.

4. Poll voting is tied to `collab_edit` – but voting isn't really editing. Either give it its own permission or at least hardcode it to `trip_member`.

**Minor:**

- `NoteFormModal` props were changed from required to optional without an obvious reason – this can hide bugs, please revert if not needed.
- Double-check the i18n keys for completeness across all 12 language files.

slashwarm:
Danke, für deine Arbeit!

1. Yes, they're protected by the router-level middleware.
2. You are right, the only reason I did that is to allow the Demo User in demo mode to edit the default shared event. I will promptly push a commit to set the default to trip_owner. Let me know if you want some custom functionality for demo mode.
3. Done
4.  Done
5. Done
6. Done

mauriceboe:
Looks good now, merged. Thanks for the quick fixes and the solid work!

PR #240, opened 2026-03-31
feat: added all supported currencies from exchangerate-api (#229)

Summerfeeling:
This change add all supported currencies of the API that is already in use and implements the request #229.


mauriceboe:
Clean addition, merged into dev. Thanks!

PR #242, opened 2026-03-31
Update type defs to support suggestion items + dedupe keys

Stew8260:
This PR fixes some errors inside the translation files. 

- Updated TypeScript types from Record<string, string> to support structured suggestion items: Record<string, string | { name: string; category: string }[]>
- Ensured all language files match en.ts
- Removed duplicate translation keys across some files
- Kept the most recent definitions


mauriceboe:
Hey, the type fixes for es/fr/nl/ru/zh are good, but some of the removed keys are still actively used in AtlasPage.tsx:

- `atlas.year` — used as placeholder in the year selector (lines 808, 1024)
- `atlas.bucketNamePlaceholder` — used as placeholder in the bucket list input (line 983)

Removing these would cause raw key strings to show up in the UI instead of translations. Please keep those keys and resubmit with only the actual duplicates and the type fixes.

Issue #218, opened 2026-03-31
[BUG] Trip tab very laggy with 60+ places [HELP WELCOME!]


lucaam:
**Describe the bug**
I am hosting this fantastic tool in my k8s cluster using the provided helm chart but I this this the plan tab is very laggy when dealing with around 60 places.
That's not the case if I just create a new trip in the same instance and I test with 5/10 places only.

The laggy pieces are:
Map (pan/zoom)
Day planning side panel (even to just scroll)
Places side panel (drag and drop places)

**To Reproduce**
1. Create a trip
2. Add or import more than 50 places
3. Navigate to the plan tab and play around with places and day planning

**Expected behavior**
Smooth environment even with 50+ places

**Desktop (please complete the following information):**
 - OS: Ubunut 24
 - Browser Edge
 - Version 2.7.1

mauriceboe:
Thanks for the detailed report! I've actually noticed the same behavior myself — the Plan tab starts to feel sluggish once you get past ~50 places, especially the map interactions and drag-and-drop. It's been on my radar and I'll be taking another closer look at it today to see where the main bottlenecks are.

My initial suspicion is that it's a combination of too many re-renders. I'll keep this issue updated with any findings.
Appreciate you taking the time to report this — and glad you're enjoying TREK!

jubnl:
I'm currently working on this issue, and yeah there's way too much rerenders. I currently have a fix the for map and the places panel and I'm searching a proper fix for the drag and drop for days assignments. I'll make a pr once I have everything running smooth with 50+ places

018929014:
Also encountered the issue and seems to me related to JPEG decode per PlaceAvatar, my understanding was that the pictures we retrieve from wiki are way too big and the browser struggle to render it. I did some hacky fix on https://github.com/018929014/TREK, but didn't really had time to look at it. If that can help somehow.

lucaam:
Thanks guys, I was also having a look but it seems to be harder than expected, on my dev env tha map is quite smooth now but the 2 side panels are very laggy.

mauriceboe:
For me, it was mainly the list of locations that was really laggy. @jubnl I'll leave this performance bug out for now and take care of the rest later; maybe you'll find a clean solution. Thanks! 🙏

jubnl:
I mounted an instances with some of the fixes but that's kinda hard for me to know if that's really fixed. Could you guys please test a bit the perfs ? I already created a trip and added 71 places:

[...]

PR #243, opened 2026-03-31
Performance on trip planner

jubnl:
Added caching, memoization and async decoding for place avatars

related to https://github.com/mauriceboe/TREK/issues/218


mauriceboe:
Solid performance improvements across the board. Merged into dev with our existing icon rendering fix preserved. The Wikipedia thumbnail change alone should make a noticeable difference. Thanks!

PR #248, opened 2026-03-31
fix: require authentication for trip file downloads

ktorq:
## Summary

- Adds JWT authentication middleware to the `/uploads/files/*` route so trip documents can no longer be accessed without a valid token
- After authentication, verifies the user has access to the trip that owns the file via `canAccessTrip()`
- Avatars, covers, and photos remain publicly accessible (non-sensitive, used on login/sharing screens)

## Details

Previously the `/uploads/:type/:filename` handler relied on UUID unguessability as the sole access control for trip files (documents, PDFs, spreadsheets, etc.). UUIDs can leak via Referer headers, logs, browser history, or shared screens, making this security-by-obscurity insufficient for potentially sensitive travel documents.

The fix conditionally applies the existing `authenticate` middleware only when `type === 'files'`, then looks up the file's `trip_id` in `trip_files` and calls `canAccessTrip()` to verify the requesting user is the trip owner or a member.

### DIFF
--- a/server/src/index.ts
+++ b/server/src/index.ts
@@ -117,8 +117,19 @@ import { authenticate } from './middleware/auth';
 app.use('/uploads/avatars', express.static(path.join(__dirname, '../uploads/avatars')));
 app.use('/uploads/covers', express.static(path.join(__dirname, '../uploads/covers')));
 
-// Serve uploaded files (UUIDs are unguessable, path traversal protected)
-app.get('/uploads/:type/:filename', (req: Request, res: Response) => {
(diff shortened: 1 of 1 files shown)


mauriceboe:
Hey, thanks for looking into this – but I already solved this on dev today. Direct access to `/uploads/files` is blocked entirely (returns 401), and file downloads go through an authenticated endpoint at `/api/trips/:tripId/files/:id/download` with JWT token verification + trip access check.

Closing this since it's already covered.

ktorq:
Good to hear, thanks!

PR #254, opened 2026-04-01
fix: force IPv4 for weather API requests to prevent ETIMEDOUT on Docker

shanelord01:
node-fetch v2 ignores NODE_OPTIONS --dns-result-order=ipv4first because it creates its own HTTPS agent. On Docker bridge networks where IPv6 cannot route to the internet, every outbound request hangs until TCP timeout.

- Add custom https.Agent({ family: 4 }) passed to all open-meteo fetch calls
- Add NODE_OPTIONS=--dns-result-order=ipv4first to docker-compose as defence in depth

Replacing PR https://github.com/mauriceboe/TREK/pull/180


shanelord01:
Incorrect PR - irrelevant commits included.

PR #257, opened 2026-04-01
Security hardening, encryption at rest

jubnl:
## Summary

This PR delivers a significant security hardening pass

### Security
- **Encryption at rest**: SMTP password, OIDC client secret, and Immich API key are now encrypted with AES-256-GCM; `ENCRYPTION_KEY` is required at startup with a migration script and backwards-compatible fallback for existing deployments
- **JWT hardening**: Decoupled JWT secret from the encryption key; removed `JWT_SECRET` env var (server manages it exclusively); replaced JWT tokens in URL query params with short-lived ephemeral tokens; migrated JWT storage from `localStorage` to `httpOnly` cookies (`trek_session`)
- **OIDC**: Fixed redirect URI construction to ignore untrusted `X-Forwarded-Host`; encrypted client secret at rest
- **Rate limiting**: Added independent rate limiter for MFA verification endpoints
- **SSRF protection**: Added guard for link preview and Immich URL fetches
- **Input validation**: Fixed ICS header injection in calendar export; enforced consistent password policy across all auth flows; restricted trip listing and access to own/shared trips only
- **Backup restore**: Validate uploaded backup DB before restoring
- **DB reliability**: Wrapped each migration in a transaction and surfaced previously swallowed errors; routed db helper functions through the null-safe proxy

### Bug Fixes
- Unified password validation error to show all requirements at once
- Password policy consistently enforced across registration, OIDC provisioning, and admin flows

## Test plan
- [x] confirm migration script re-encrypts existing secrets
- [x] Confirm SMTP, OIDC, and Immich credentials are stored encrypted in the DB
- [x] Test login/logout with `trek_session` httpOnly cookie; verify token is not accessible via JS
- [x] Validate MFA rate limiter blocks brute-force attempts
- [x] Test SSRF guard rejects private/local URLs for link preview and Immich
- [x] Confirm calendar `.ics` export rejects header-injection payloads
- [x] Confirm password validation shows all failing requirements at once on registration and password change

### DIFF
--- a/server/src/routes/immich.ts
+++ b/server/src/routes/immich.ts
@@ -92,7 +92,8 @@ router.get('/status', authenticate, async (req: Request, res: Response) => {
 router.post('/test', authenticate, async (req: Request, res: Response) => {
   const { immich_url, immich_api_key } = req.body;
   if (!immich_url || !immich_api_key) return res.json({ connected: false, error: 'URL and API key required' });
-  if (!isValidImmichUrl(immich_url)) return res.json({ connected: false, error: 'Invalid Immich URL' });
+  const ssrf = await checkSsrf(immich_url);
+  if (!ssrf.allowed) return res.json({ connected: false, error: ssrf.error ?? 'Invalid Immich URL' });
   try {
     const resp = await fetch(`${immich_url}/api/users/me`, {
       headers: { 'x-api-key': immich_api_key, 'Accept': 'application/json' },


mauriceboe:
Great work on this PR — the separation of encryption key from JWT secret, SSRF guard, ephemeral tokens, and RCE removal are all solid.

One concern: `ENCRYPTION_KEY` being required with `process.exit(1)` if not set is a breaking change for existing installations. A simple `git pull` would result in a non-starting server.

Could this be handled with automatic fallback instead?

1. Read `ENCRYPTION_KEY` from env
2. If not set, check if `data/.jwt_secret` exists (existing install) and use that as the encryption key
3. If neither exists (fresh install), auto-generate and persist to `data/.encryption_key`

This way existing installs keep working without manual intervention, encrypted data stays readable (same key), and the security benefit (separate encryption key from JWT signing secret) is preserved. Users who want to explicitly set their own key via env can still do so.

jubnl:
PR is now ready. I also moved JWT token from local storage to a proper http only cookie. I changed the PR body to reflect all that.

jubnl:
Resolving conflicts

mauriceboe:
@jubnl how long would that take you? I’d like to keep working on the dev branch right away, and to avoid any conflicts later on, I’d be happy to merge your PR.

jubnl:
2 mins

jubnl:
Done!

PR #261, opened 2026-04-01
fix: remove node-fetch, use native fetch (Node 22)

shanelord01:
### Problem

`node-fetch@2.7.0` pulls in `whatwg-url@5.0.0` → `tr46@0.0.3`, which imports the built-in `punycode` module deprecated since Node.js 21+. This produces the following warning on every server start:

```
(node:1) [DEP0040] DeprecationWarning: The `punycode` module is deprecated. Please use a userland alternative instead.
```

### Fix

Removed `node-fetch` entirely and switched to Node 22's native global `fetch`. This eliminates the entire dependency chain (`node-fetch`, `whatwg-url`, `tr46`, `webidl-conversions`).

### Changes

- Removed `node-fetch` from `package.json` and `package-lock.json`
- Removed `import fetch from 'node-fetch'` in 6 files: `places.ts`, `oidc.ts`, `auth.ts`, `maps.ts`, `weather.ts`, `notifications.ts`
- Replaced `require('node-fetch')` in `collab.ts` with global `fetch`
- In `weather.ts`, added `dns.setDefaultResultOrder('ipv4first')` to preserve IPv4-first behaviour for Docker bridge networks where IPv6 cannot route to the internet (replaces the approach in PR #180 which used a custom `https.Agent({ family: 4 })`)

### Compatibility audit

All 7 source files use only standard fetch features (`json()`, `text()`, `ok`, `status`, `signal`, `redirect`, `headers`). No node-fetch-specific classes (`Response`, `Request`, `Headers`) or stream handling were in use. `redirect: 'error'` confirmed compatible with native `fetch`.

### Supersedes #180

Using Claude Code - please ignore or use - it's up to you.

### DIFF
--- a/server/package-lock.json
+++ b/server/package-lock.json
@@ -19,7 +19,6 @@
         "jsonwebtoken": "^9.0.2",
         "multer": "^2.1.1",
         "node-cron": "^4.2.1",
-        "node-fetch": "^2.7.0",
         "nodemailer": "^8.0.4",
         "otplib": "^12.0.1",
         "qrcode": "^1.5.4",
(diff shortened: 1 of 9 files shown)


mauriceboe (changes requested):
Hey, thanks for the clean PR – solid work removing `node-fetch` in favor of native fetch, and the compatibility audit in the description is appreciated.

I switched the base branch to `dev` (that's where we merge feature/fix branches), but there are merge conflicts. Could you rebase onto `dev` so we can get this merged?

One minor note: `dns.setDefaultResultOrder('ipv4first')` in `weather.ts` is a process-wide side effect – it affects all DNS lookups, not just weather requests. Might be worth either moving it to the server entrypoint or at least adding a comment that it's global. Not a blocker though.

jubnl:
Node fetch as been removed in dev

PR #270, opened 2026-04-01
feat: copy/duplicate trips from dashboard

slashwarm:
Adds a copy button to all dashboard card types so users can duplicate existing trips.

**What it does:**
- New `POST /api/trips/:id/copy` endpoint that deep copies all planning data (days, places, assignments, reservations, budget items, packing lists + bags, accommodations, day notes) with full FK remapping in a single transaction
- Copy button on spotlight, grid, list, and archived cards
- Gated by `trip_create` permission
- Copied trip title gets a localized suffix (e.g. "My Trip (copy)", "Utazásom (másolat)")
- Packing items have `checked` reset to 0, `paid_by_user_id` on budget items is cleared

**Intentionally skipped:**
- Files (trip_files, file_links, photos)
- Collab data (notes, polls, messages)
- Members, share tokens, per-user assignments

Also added `reminder_days` to the `Trip` interface, removing `as any` casts in the PUT route.


mauriceboe:
Hey, thanks for the great contribution! Really clean implementation – the transaction-based deep copy with FK remapping across 11 tables is solid work. Merged it into dev.

Small note for a follow-up: the budget items copy doesn't include `expense_date`, so that field gets lost on copy. Quick fix if you want to grab it.

PR #271, opened 2026-04-01
Add category icons and simplify Places sidebar

entropyst72:
Show category icons in place/category select controls and streamline sidebar filtering/UI.


mauriceboe:
Hey, thanks for the contribution! The category icons in the dropdowns are a nice touch.

However I can't merge this as-is for a couple of reasons:

1. **Google Maps URL resolver removed** – the block in `PlaceFormModal.tsx` that resolves Google Maps URLs was deleted. That's an unrelated feature removal and would break existing functionality.

2. **Multi-category filtering downgraded to single-select** – the sidebar filter previously allowed selecting multiple categories at once, this change reduces it to one at a time. That's a UX downgrade.

If you'd like to rework this to just add the category icons without the other changes, I'd be happy to take another look!

PR #315, opened 2026-04-02
fix: badge number partially hidden

marco783:
In the trip map, places in the day with a photo taken from google maps, have the badge partially hidden

This small fixes resolve the issue.


marco783:
fixed by mauriceboe in this exact moment :D

PR #334, opened 2026-04-03
Add Accomodation to PDF

micro92:
Screenshot from 2026-04-02 20-43-55

Display Accomodation linked to the specific day.

### DIFF
--- a/client/src/components/PDF/TripPDF.tsx
+++ b/client/src/components/PDF/TripPDF.tsx
@@ -126,8 +126,8 @@ export async function downloadTripPDF({ trip, days, places, assignments, categor
   const sorted = [...(days || [])].sort((a, b) => a.day_number - b.day_number)
   const range = longDateRange(sorted, loc)
   const coverImg = safeImg(trip?.cover_image)
-  //retrieve accomodations for the trip to display on the day sections and prefetch their photos if needed
-  const accomodations = await accommodationsApi.list(trip.id);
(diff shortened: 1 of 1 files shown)


micro92:
relates to Feature request #333

mauriceboe (changes requested):
Hey, nice feature – showing accommodations on the PDF day sections is a great addition, and the check-in/check-out distinction makes a lot of sense.

One thing to address before this can be merged:

### No emojis – use Lucide icons instead

TREK has a no-emoji policy throughout the UI. All icons come from [Lucide React](https://lucide.dev/icons/) (`lucide-react`), which is already imported at the top of `TripPDF.tsx`. You're currently using 🛎️🧳🏨📌📝🔑 inline – these need to be replaced with the appropriate Lucide icons rendered as SVG strings for the PDF HTML.

The existing code in the same file already does this for place cards (check how `getCategoryIcon` and `NOTE_ICON_MAP` work). For your use case, relevant Lucide icons would be something like:
- `Hotel` or `BedDouble` → accommodation
- `LogIn` / `LogOut` → check-in / check-out
- `MapPin` → address
- `StickyNote` or `FileText` → notes
- `KeyRound` → confirmation code

Everything else looks good – the filtering logic, the layout with single/multi accommodation cards, and the HTML/CSS structure all work well. Looking forward to the update!

micro92:
Emoji Replaced by Lucide Icons

image

PR #337, opened 2026-04-03
feat: external S3-compatible storage backends

swheettaos:
## External Storage Backend Support

This PR adds configurable S3-compatible external storage targets for backups, photos, trip files, covers, and avatars.

### What's new

**Admin Panel**
- New "Storage Targets" panel — create, test, and delete named S3-compatible storage targets (AWS S3, MinIO, Wasabi, Backblaze B2, Cloudflare R2, DigitalOcean Spaces, and any S3-compatible service)
- New "Storage Assignments" panel — independently assign each storage purpose (backups, photos, files, covers, avatars) to any configured target
- Storage credentials are stored in a dedicated `storage_targets` table and never exposed back to the client after save

**Backend**
- Pluggable `StorageBackend` interface (`local`, `s3`, `encrypted`) with a factory/cache layer
- Optional AES-256-GCM encryption wrapper (`encrypted.ts`) — encrypts on upload, decrypts on stream
- First-boot environment variable bootstrap (`STORAGE_S3_*`) for automated/Helm deployments
- Local filesystem remains the default; nothing breaks if no external target is configured
- Fallback to local filesystem for avatars/covers if a target is misconfigured, preventing site-wide media breakage

**Helm / Docker Compose**
- New `Secret` entries for S3 credentials
- `ConfigMap` additions for non-sensitive storage config
- Docs with full examples for Docker Compose, Kubernetes/Helm, and migration from local storage

### What was considered and not implemented

**NFS/network filesystem support** was explored but dropped. The available Node.js NFS client libraries are either unmaintained or rely on outdated native dependencies with CVEs. Writing a custom wrapper around the low-level libraries was beyond the scope of this effort, so NFS is not supported — S3-compatible object storage covers the main use cases anyway.

### Translations

Translation strings were added for all 12 supported locales (`en`, `de`, `fr`, `es`, `it`, `nl`, `pt`, `ar`, `cs`, `hu`, `ru`, `zh`). **These translations were generated with AI assistance.** I cannot vouch for their accuracy or natural phrasing — native speaker review would be very welcome before merging to main.

### Security

Several rounds of AI-assisted security audit were performed on this feature. Key areas reviewed and hardened:
- S3 credentials stored in a `Secret` (not `ConfigMap`) in Kubernetes; never returned to the client after initial save
- Encryption key derivation uses PBKDF2 with per-file random IVs (AES-256-GCM)
- Filename sanitisation on all storage paths to prevent path traversal
- Backup filename validation regex to prevent directory traversal on restore
- SQLite foreign-key constraints enforced at connection time

I tried my best to make this as secure as possible, but I'd appreciate a thorough review of the storage and encryption code in particular.

### Bug fixes included

- `admin.ts`: Fixed `writeAudit` call in `rotate-jwt-secret` — `userId` was being passed as `null`
- `collabService.ts`: Removed stale `fs.unlinkSync` loops left over from before the storage abstraction
- `encrypted.ts`: Stream error propagation fixed; temp files now use `data/tmp` instead of `os.tmpdir()`
- `factory.ts`: Cache key collision fixed (numeric vs string key mismatch)


swheettaos:
ok.. I tried on this one.. I had to refactor like 4 times because you have been a mad man with commits on your dev branch.  I really tried to keep up .. in the end I used a bunch of AI to help so they may be some slop .. and if you don't like it just throw it away.  I think it's pretty good.. but I'm sure I'm still missing stuff.

mauriceboe:
Hey, I appreciate the effort that went into this — the scope is ambitious and the storage abstraction is well-structured.

However, I can't merge this. A few reasons:

1. **The AI prompt file (`plan-trekExternalStorage.prompt.md`) was committed to the repo.** This is the full implementation plan you fed to an AI to generate this code. That's not something that should end up in a production codebase.

2. **"if you don't like it just throw it away" is not the level of confidence I need for a PR that introduces encryption, credential storage, and a full storage abstraction layer.** This is security-critical code — I need contributors to stand behind what they submit, not treat it as disposable.

3. **6000+ lines of self-admittedly AI-generated code with "some slop"** touching backups, file handling, and encryption is too risky to merge without significantly more scrutiny and testing than what was done here.

The feature itself (S3 storage backends) is something I'd welcome in the future, but it needs to come from someone who can vouch for the code they're submitting.

swheettaos:
The plan was submitted by accident and was primary (at least in the beginning) used to help me audit my code and fill gaps for things like translations and security audit, and to ensure I wasn’t missing features.   That being said it is massive and you are correct, while it works I’m not confident enough to say my approach is the correct one. Maybe one of the other devs can run with this. I understand completely.

PR #344, opened 2026-04-03
added trip member count to dashboard

marco783:
Added the count of the member of the trip to the dashboard page.

Added a new string to the translations `dashboard.members` and translated with AI in the various languages, so, there could be errors :D

immagine


jubnl:
lgtm!

PR #348, opened 2026-04-03
added map preview to settings, change latitude and longitude with lef…

marco783:
Added map preview to settings page.
Also, clicking on the map, the latitude and longitude will be set.

immagine


mauriceboe:
Hey, thanks for the contribution – nice addition to the settings page! Merged.

PR #376, opened 2026-04-03
chore(helm): add config/secret checksum to deployment

darioackermann:
This PR adds a config and secret checksum to pod template annotations of the spec of the deployment of the Helm Chart.

This is useful in cases where GitOps is used (e.g., ArgoCD), because Kubernetes does not automatically redeploy pods when underlying resources (here: configmap, secrets) are changed.


jubnl:
Thanks for the PR!

PR #477, opened 2026-04-06
feat(atlas): show visited cities on atlas map

bigjoky:
## Summary
- load visited cities into the Atlas page
- keep regions visible on zoom while adding city markers as a complementary layer
- improve city marker contrast and stacking so they stay readable above regions

## Verification
- npm run build


jubnl:
https://github.com/mauriceboe/TREK/blob/main/CONTRIBUTING.md

PR #488, opened 2026-04-07
feat: Import Places / Activities From Keyhole Markdown (KMZ/KML)

biasutti:
**See points to discuss**

### Feature summary:
- Add fileupload dialog to import places from .kmz and .kml files. This open format is used by Google MyMaps and Google Earth etc.
- Parse provided folders and match to existing categories for places.

### Tests:
- Implemented integration tests
- Conducted local testing with provided test files in test fixture folder

### Points to discuss / maybe further merge requests:
1. Button placement in sidebar. Currently I added a third option between GPX and Google Lists. This causes problems with the length of the label KMZ / KML and further extensibillity of the sidebar. Any suggestions / wishes on label name or button placement.

grafik

2. The implementation in the frontend is prepared to support further formats. A GPX integration into the same dialog would also be possible. This would free up some space in the sidebar. What are your thought on this?

3. Not all imported data can be matched to some places on Google maps. This causes some 404 api errors when loading images for those place. I have to look further into it what a solution could look like.

----

Update to discussion-point 3.:
The picture cant be found for some places as the KMZ import does not provide a Google PlaceID and therefore the picture search algo is using a fallback with coordinates which not always returns a picture.
Currently for every unmatched place there is a warning in the log:
`[WARN] 2026-04-07T19:31:17 GET /place-photo/coords%3A34.951474%3A135.7923898 404 1ms ip=::1`
Are we okay with this warning as its not directly caused by this change or should I improve this?


jubnl:
Closed in favor of #663

PR #497, opened 2026-04-07
fix: synology server url

tiquis0290:
change of how synology plugin expects server url

should fix problems connected to unabling to connect to server


jubnl:
Please open your pr again with target to dev. Tests must be run properly, pull dev into your branch and reopen a new pr with the right branch as target.

PR #499, opened 2026-04-07
fix; synology url

tiquis0290:
change in url expected by trek from user


jubnl:
@tiquis0290 is your patch fixing issue #500 ?

tiquis0290:
Should fix it, also i have update login method to use v6 instead of v3 and added support for OTP (MFA)

jubnl:
Oh and I just checked and you did not included the tests that have been pushed to the dev branch. Please open a PR from tiquis0290/TREK/tree/dev (the branch should be up to date with the current mauriceboe/TREK dev branch) to mauriceboae/TREK/tree/dev.

tiquis0290:
ok

MikeDabrowski:
This change does not allow synology photos aliased to `photos` for example - Is this intentional? 
My setup is fairly new so was not a problem to change alias but not sure for others with more mature setups?

PR #503, opened 2026-04-07
fix synology

tiquis0290:
Changing how synology threats endpoint

Changing auth method in synology from v3 to v6 and adding OTP (MFA) support


tiquis0290:
because of pr i made to get the changes to the branch befor it got siconected hopefully now its all as required

server/src/services/memories/synologyService.ts
```diff
@@ -303,6 +337,17 @@ export async function updateSynologySettings(userId: number, synologyUrl: string
     }
 
     _clearSynologySID(userId);
+    console.log(synologyOtp, synologyOtp?.trim(), synologyOtp && synologyOtp.trim()); //debug log to verify OTP presence
```
jubnl: OTP should not leak in the logs

jubnl (changes requested):
Since you normalize the urls in the DB, could you please add this normalization part  when you save the base url ? or just add a warning in the frontend saying something along the line of "Are you sure this normal that this url does not end with /photo"
And in the user settings, please add a bit of documentation regarding sonology (sonology server requirements, so DSM version and so on)

tiquis0290:
i normalize url each time iam calling it, i need it to append /photo to old urls in db

so dont see necessary to use it when storing new ones

sorry i missed the console log

jubnl:
Closed in favor of #565

PR #528, opened 2026-04-08
DRAFT: MCP Token query parameter support

MB175:
This pull request adds support for authenticating MCP API requests using a token provided as a URL query parameter (`?token=...`), in addition to the existing `Authorization` header method. 

This makes it easier for clients that cannot set custom headers to authenticate. A new integration tests are added to verify the behavior.

Copilot was used to quickly identify where the validation of the MCP-Token is taking place.


MB175:
Moved into draft state until proper announcement on discord

PR #532, opened 2026-04-09
docs: Clarify FORCE_HTTPS and TRUST_PROXY configuration

### LINKED ISSUE #531: Clarify FORCE_HTTPS and TRUST_PROXY configuration documentation
## Problem

The documentation for `FORCE_HTTPS` and `TRUST_PROXY` environment variables is not clear enough for users, leading to common configuration issues where:

1. Users set `FORCE_HTTPS=true` in `docker-compose.yml` but access TREK directly on `http://host:3000`
2. This causes all HTTP requests to be redirected to HTTPS, but since there's no TLS termination, the redirects fail
3. Users cannot access their TREK instance and experience confusing infinite redirect loops

## Current Documentation Issues

- `FORCE_HTTPS` documentation doesn't clearly indicate when to use `true` vs `false`
- No warning about the consequences of setting `FORCE_HTTPS=true` with direct port access
- `TRUST_PROXY` documentation doesn't emphasize it's only needed when actually behind a reverse proxy
- Users might not understand the relationship between these two settings

## Proposed Solution

Update the documentation to:
1. Explicitly warn that `FORCE_HTTPS=false` is required for direct access
2. Clearly separate the use cases for direct access vs reverse proxy setups
3. Add examples showing both configurations

## Additional Context

This issue affects new users particularly, as they might start with the `docker-compose.yml` example and try to access the app directly, not understanding they need a reverse proxy for `FORCE_HTTPS=true`.

## Impact

- **High**: Prevents application access for new users
- **Medium**: Causes confusion and support requests
- **Low**: Easy fix with better documentation

## Related Files

- README.md (lines 288-290): Environment variables table
- README.md (around line 176): Docker Compose example section

luojiyin1987:
## Summary

This PR improves the documentation for `FORCE_HTTPS` and `TRUST_PROXY` environment variables to prevent common configuration issues that cause infinite redirect loops.

### README.md
1. **FORCE_HTTPS documentation** (line 288):
   - Added explicit warning: "If you access TREK directly on `http://host:3000`, keep this `false`."
   - Clear guidance for direct access scenarios

2. **TRUST_PROXY documentation** (line 290):
   - Clarified that this should only be used "when TREK is actually behind a reverse proxy"
   - Prevents misconfiguration when not using a reverse proxy

3. **Docker Compose example** (around line 176):
   - Added warning: "If you access TREK directly on `http://<host>:3000` without nginx, Caddy, Traefik, or another TLS-terminating proxy in front of it, set `FORCE_HTTPS=false` and remove `TRUST_PROXY`"

## Problem Solved

This fixes the issue where:
- Users set `FORCE_HTTPS=true` but access directly via `http://host:3000`
- The app redirects all requests to HTTPS, causing connection failures
- Users experience confusing behavior and cannot access their TREK instance

## Testing

The changes are documentation-only and don't affect code behavior. The documentation now clearly explains:
- When to use `FORCE_HTTPS=true` (with reverse proxy)
- When to use `FORCE_HTTPS=false` (direct access)
- When to use `TRUST_PROXY` (only when behind a reverse proxy)

## Related Issue

Closes #531

### Direct Access (Recommended for beginners)
```yaml
environment:
  FORCE_HTTPS: false
  # TRUST_PROXY not needed
```

### Reverse Proxy (Production)
```yaml
environment:
  FORCE_HTTPS: true
  TRUST_PROXY: 1
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -174,6 +174,8 @@ services:
       start_period: 15s
 ```
 
+This example is aimed at reverse-proxy deployments. If you access TREK directly on `http://<host>:3000` without nginx, Caddy, Traefik, or another TLS-terminating proxy in front of it, set `FORCE_HTTPS=false` and remove `TRUST_PROXY` to avoid redirects to a non-existent HTTPS endpoint.
+
 ```bash
 docker compose up -d
 ```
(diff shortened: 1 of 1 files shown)


jubnl:
Thank you very much!

Colcothar:
Hi @jubnl , @luojiyin1987 

Firstly, thanks for the great app.

Is the intended behaviour of FORCE_HTTPS=true correct when using reverse proxies?

In standard setups with reverse proxies (e.g. NGINX or Traefik), TLS is terminated at the proxy and the app receives HTTP with X-Forwarded-Proto headers. In these cases, it’s unusual to also enforce HTTPS at the application layer, and it can sometimes lead to redirect or proxy conflicts depending on configuration.

It might be worth clarifying this in the docs, as the current examples suggest using FORCE_HTTPS=true in reverse-proxy deployments, which could be misleading.

luojiyin1987:
@Colcothar  Thanks.  I created Issue and  pull request.

#560

PR #542, opened 2026-04-09
feat: add autocomplete/typeahead to place search

ben-haas:
## Description
Add real-time autocomplete/typeahead suggestions to the place search field in PlaceFormModal. When users type in the search box, they now see a dropdown of place suggestions powered by Google Places Autocomplete API (with Nominatim fallback for users without a Google API key). Selecting a suggestion fetches full place details and populates the form.

Key changes:
  - New POST /api/maps/autocomplete backend route and autocompletePlaces service function
  - Debounced autocomplete dropdown in PlaceFormModal with full keyboard navigation (arrow keys, enter, escape) and mouse support
  - Location bias derived from existing trip places for more relevant suggestions
  
## Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update

## Checklist


client/src/components/Planner/PlaceFormModal.tsx
```diff
@@ -25,6 +25,8 @@ interface PlaceFormData {
   website: string
 }
 
+const GOOGLE_MAPS_URL_RE = /^https?:\/\/(www\.)?(google\.[a-z.]+\/maps|maps\.google\.[a-z.]+|maps\.app\.goo\.gl|goo\.gl)/i
```
jubnl: This regex is flawed: https://regexr.com/8ljek

jubnl (changes requested):

ben-haas:
I forgot to push up the tests. I'll look into the regex as well.

jubnl:
oof.. The failing test has nothing to do with your pr, don't worry. I'll fix it when I have some time.

Regarding the regex, it still has some issues. By checking `https://google.[2-3 chars].[2 chars]/maps`, this work : `https://google[.]aaa[.]ch/maps`

I've updated the test case: https://regexr.com/8ljek 

An idea you can explore is instead to URL(urlString.trim()), then use the hostname prop from the URL object

ben-haas:
> oof.. The failing test has nothing to do with your pr, don't worry. I'll fix it when I have some time.
> 
> Regarding the regex, it still has some issues. By checking `https://google.[2-3 chars].[2 chars]/maps`, this work : `https://google[.]aaa[.]ch/maps`
> 
> I've updated the test case: https://regexr.com/8ljek
> 
> An idea you can explore is instead to URL(urlString.trim()), then use the hostname prop from the URL object

Ya, that seems cleaner. Just pushed an update.

solarssk:
Hey! One question about the location bias logic - currently I understand, it picks the first place with coordinates from the trip:

What happens if the trip spans multiple countries (e.g. Japan + China)? The bias will always point to whichever country was added first, which could result in poor suggestions when searching for places in the other country.

ben-haas:
> Hey! One question about the location bias logic - currently I understand, it picks the first place with coordinates from the trip:
> 
> What happens if the trip spans multiple countries (e.g. Japan + China)? The bias will always point to whichever country was added first, which could result in poor suggestions when searching for places in the other country.

That's a good point! It looks like a [bounding box](https://developers.google.com/maps/documentation/places/web-service/place-autocomplete#location-bias-restriction) can be used in the locationBias parameter, which would be an improvement, however I'm re-thinking this feature now. There is a lot of nuance to trips that influences how the bias would be useful. For example, if I was planning a road trip, a search bias that included a corridor around my route would be nice, but if I was flying across a country or continent, obviously there is no use in having a search bias along the route. 

I think the bias parameter should be removed entirely for now and possibly revisited later. There is more complexity than I originally considered and I'm not sure that it adds that much value.

What do you think @solarssk?

ben-haas:
I caught a bug and made some changes to help with the searchBias. I also merged in dev and resolved all conflicts. Here is a summary:

Autocomplete search with location bias

- Location bias uses a bounding box derived from the trip's existing places (or the map viewport) to improve suggestion relevance, with validation to reject degenerate/zero-area bounds

OSM place details fix

  - Bug fix: When no Google Places API key is present, selecting an autocomplete suggestion would not populate the modal fields (name, address, coordinates, website). Root cause: the Overpass API detail lookup only returned metadata (phone, website, hours) but not core fields. Added a Nominatim /lookup call (run in parallel with Overpass) to fetch name, address, lat, and lng for OSM places.

  Loading UI

  - When a suggestion is selected, the place name is populated in the Name field immediately from the suggestion text
  - An inline spinner appears beside the name while the remaining details (address, coordinates, website) are fetched
  - hasMapsKey now updates reactively when the API key is saved or removed, so the "Search via OpenStreetMap" hint appears/disappears without a page refresh

  Other

  - Added places.loadingDetails translation key across all 14 locales
  - Added tests for mapsApi.autocomplete and autocompletePlaces service interactions

jubnl:
Closed in favor of #661

PR #549, opened 2026-04-10
README.md: doc update, nginx reverse proxy conf

icedieler:
When using nginx as a reverse proxy it might refuse file uploads because the default upload size is 1m only. TREK allows file sizes of up to 50m.

Thus set the `client_max_body_size` to `50m` in the nginx configuration.


jubnl:
https://github.com/mauriceboe/TREK/commit/04c58e6e0fe0f0d06ad55f40eafc5815d989aafc

PR #585, opened 2026-04-12
fix: resolve Atlas US state codes as US

brdweb:
## Summary
- treat US postal abbreviations in Atlas place addresses as `US` instead of foreign country codes
- prefer cached `place_regions.country_code` before re-parsing ambiguous addresses

## Validation
- `TMPDIR=/tmp TEMP=/tmp TMP=/tmp PATH=/run/user/1000/fnm_multishells/211616_1775960379294/bin:$PATH npm test -- tests/integration/atlas.test.ts`
- `TMPDIR=/tmp TEMP=/tmp TMP=/tmp PATH=/run/user/1000/fnm_multishells/211616_1775960379294/bin:$PATH npm test -- tests/unit/mcp/tools-atlas.test.ts`
- `TMPDIR=/tmp TEMP=/tmp TMP=/tmp PATH=/run/user/1000/fnm_multishells/211616_1775960379294/bin:$PATH npm test`
- `TMPDIR=/tmp TEMP=/tmp TMP=/tmp PATH=/run/user/1000/fnm_multishells/211616_1775960379294/bin:$PATH npm run build` (client)

## Context
This fix validated against imported TripIt data where addresses ending in `GA`, `CO`, `LA`, and `AR` were being bucketed as Gabon, Colombia, Laos, and Argentina on the Atlas page.

### DIFF
--- a/server/src/services/atlasService.ts
+++ b/server/src/services/atlasService.ts
@@ -108,6 +108,22 @@ export const NAME_TO_CODE: Record<string, string> = {
   'saudi arabia':'SA','saudi-arabien':'SA','albania':'AL','albanien':'AL',
 };
 
+const US_STATE_CODES = new Set([
+  'AL', 'AK', 'AZ', 'AR', 'CA', 'CO', 'CT', 'DE', 'FL', 'GA',
+  'HI', 'ID', 'IL', 'IN', 'IA', 'KS', 'KY', 'LA', 'ME', 'MD',
+  'MA', 'MI', 'MN', 'MS', 'MO', 'MT', 'NE', 'NV', 'NH', 'NJ',
+  'NM', 'NY', 'NC', 'ND', 'OH', 'OK', 'OR', 'PA', 'RI', 'SC',
+  'SD', 'TN', 'TX', 'UT', 'VT', 'VA', 'WA', 'WV', 'WI', 'WY',
+  'DC',
+]);
+
+const US_COUNTRY_MARKERS = new Set([
+  'united states',
+  'united states of america',
+  'usa',
+  'us',
+]);
+
 export const CONTINENT_MAP: Record<string, string> = {
   AF:'Africa',AL:'Europe',DZ:'Africa',AD:'Europe',AO:'Africa',AR:'South America',AM:'Asia',AU:'Oceania',AT:'Europe',AZ:'Asia',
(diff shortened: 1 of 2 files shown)


jubnl:
Targetted branch should be the dev branch, not main.
https://github.com/mauriceboe/TREK/wiki/Development-environment
https://github.com/mauriceboe/TREK/blob/main/CONTRIBUTING.md

jubnl (changes requested):
Hey @brdweb, thanks for tracking this down, the TripIt regression is real. I'd like to request changes though, because I think the current approach is going to trade a US bug for a non-US bug. 

A lot of ISO country codes collide with US state abbreviations: DE (Germany), IN (India), AL (Albania), AR (Argentina), CO (Colombia), GA (Gabon), LA (Laos), MD (Moldova), MT (Malta), NE (Niger), VA (Vatican), etc. With the new regex, "Munich, DE" and "Mumbai, IN" resolve to US instead of their actual countries, and because getCountryFromAddress returns a non-null "US", the reverse-geocode fallback never kicks in to correct it.


Requiring a US ZIP in the regex helps, but it's not bulletproof — Brazil, Mexico, Australia, and others all have "2-letter region + digits" postal formats, so it becomes whack-a-mole. 

I think the cleaner fix is to flip the resolution priority so coordinates win over the address string, which is the ambiguous one:

```ts
async function resolveCountryCode(place: Place): Promise<string | null> {
  let code = getStoredCountryCode(place.id);
  if (!code && place.lat && place.lng) {
    code = await reverseGeocodeCountry(place.lat, place.lng);
  }
  if (!code) {
    code = getCountryFromAddress(place.address);
  }
  // ...
}
```

That:
- fixes the TripIt bug (Atlanta's coords resolve to US regardless of the "GA" suffix)
- preserves "Munich, DE" → DE, "Mumbai, IN" → IN, etc.
- lets us drop US_STATE_CODES, US_COUNTRY_MARKERS, and the state regex entirely
- place_regions caches the result, so no perf hit on repeat reads

The two new integration tests are great — it'd be awesome to add a few negative ones to lock in the behavior in both directions: "Munich, DE" stays DE, "Mumbai, IN" stays IN, "Tirana, AL" stays AL.

brdweb:
No, I agree completely. I sort of rushed it to solve only my own issue. I appreciate the discussion!

jubnl:
Closing down as no requested changes have been addressed

PR #592, opened 2026-04-12
feat(login): add language dropdown, browser auto-detection and configurable default

isaiastavares:
## Summary

- Replace the language cycling button on the login page with a full dropdown showing all 14 supported languages
- Add automatic browser/OS language detection via `navigator.languages`, applied on first visit when no language is stored
- Add `DEFAULT_LANGUAGE` env var (server-side) so self-hosters can set a default language for new users; falls back to `'en'`

## Changes

- `client/src/pages/LoginPage.tsx` — language dropdown component
- `client/src/i18n/TranslationContext.tsx` — browser language detection logic
- `client/src/i18n/index.ts` — export helpers
- `client/src/store/settingsStore.ts` — default language support
- `client/src/api/client.ts` — minor update
- `server/src/config.ts`, `server/src/app.ts`, `server/.env.example` — `DEFAULT_LANGUAGE` env var

## Test plan

- [ ] Open login page → language dropdown visible with all 14 languages
- [ ] Clear `localStorage` and reload → browser language auto-selected if supported
- [ ] Set `DEFAULT_LANGUAGE=de` on server → new users get German by default
- [ ] Change language via dropdown → persists after page reload


jubnl:
What's the right one ? You opened 2 pr..

isaiastavares:
> What's the right one ? You opened 2 pr..

@jubnl The other one was supposed to only include the translation fixes, but this feature ended up being included as well. However, I've already corrected it and separated the changes for each PR. Everything is fine now.

mauriceboe (dismissed):
~Replaced by updated review below.~

mauriceboe (changes requested):
Thanks for the contribution! The feature itself is solid — language dropdown, browser detection, and server default are all great additions. However, there are a few issues that need to be addressed before this can be merged.

### 1. Use `apiClient` instead of raw `fetch()` (`client/src/api/client.ts`)

Every other API in this file uses the project's axios `apiClient`. Raw `fetch` doesn't reject on HTTP 4xx/5xx, so `r.json()` will try to parse error HTML and fail silently.

```diff
 export const configApi = {
-  getPublicConfig: (): Promise<{ defaultLanguage: string }> =>
-    fetch('/api/config').then(r => r.json()),
+  getPublicConfig: (): Promise<{ defaultLanguage: string }> =>
+    apiClient.get('/config').then(r => r.data),
 }
```

### 2. Validate `DEFAULT_LANGUAGE` on the server (`server/src/config.ts`)

An invalid value (typo, wrong format) is silently served to all clients. Validate against supported codes at startup:

```ts
const SUPPORTED_LANG_CODES = ['de','en','es','fr','hu','nl','br','cs','pl','ru','zh','zh-TW','it','ar'];
const rawDefaultLang = process.env.DEFAULT_LANGUAGE || 'en';
if (!SUPPORTED_LANG_CODES.includes(rawDefaultLang)) {
  console.warn(`DEFAULT_LANGUAGE="${rawDefaultLang}" is not supported. Falling back to "en". Supported: ${SUPPORTED_LANG_CODES.join(', ')}`);
}
export const DEFAULT_LANGUAGE = SUPPORTED_LANG_CODES.includes(rawDefaultLang) ? rawDefaultLang : 'en';
```

### 3. Case-insensitive language matching (`client/src/i18n/TranslationContext.tsx`)

Some browsers report `zh-tw` (lowercase) instead of `zh-TW`. The exact match fails, then the prefix match maps it to `zh` (Simplified Chinese) instead of `zh-TW` (Traditional). Normalize casing:

```ts
for (const lang of browserLangs) {
  const exactMatch = supported.find(s => s.toLowerCase() === lang.toLowerCase())
  if (exactMatch) return exactMatch

  if (lang.toLowerCase().startsWith('pt')) return 'br'

  const prefix = lang.split('-')[0].toLowerCase()
  const prefixMatch = supported.find(s => s.toLowerCase() === prefix)
  if (prefixMatch) return prefixMatch
}
```

---

Please fix these and I'll re-review. The core feature is good!

isaiastavares:
@mauriceboe I've made the corrections you requested and also removed some duplicated keys.

Heads up: the test pipeline is failing, but it doesn't appear to be related to my changes — the same tests are already failing on the `dev` branch. Could you take a look when you have a chance?

jubnl:
closed in favor of #659

Issue #689, opened 2026-04-16
[BUG] [v3-pre47] synology doesnt display albums shared from other users, doesnt load photo assets added from othere users to owners albums


tiquis0290:
### Pre-flight checklist

- [x] I have searched [existing issues](https://github.com/mauriceboe/TREK/issues) and this bug has not been reported yet
- [x] I am running the latest available version of TREK

### TREK version

3.0.0.pre-47

### Describe the bug

synology doesnt display albums shared from other users, doesnt load photo assets added from othere users to owners albums

### Steps to reproduce

have shared album -> it will not be displayed

have your album where another user added photos -> asset will not load

### Expected behavior

albums should be visible

assets should load

### Deployment method

Docker Compose

### Accessing TREK from

Desktop browser

### Relevant logs or error output

```shell

```

jubnl:
see #693

tiquis0290:
> see [#693](https://github.com/mauriceboe/TREK/pull/693)

Fix doesnt store passphrases with corresponding images so after adding them they don't load

can't reopen issue but i don't think so its necessary to create a new one for it

tiquis0290:
When selecting across multiple albums the passphrase is set based on the last album selected, which leads to errors, that can even be fixed by re adding photos because they are not deleted from db when removed so they became stuck with bad passphrase @jubnl

jubnl:
@tiquis0290 
https://github.com/mauriceboe/TREK/commit/8cd5aa0d234d3bf546516ce3f6c7080d6c495f32

Issue #688, opened 2026-04-16
[BUG] [v3-pre47] pagination on synology search calculates offset from hardcoded 100 instead of supplied size


tiquis0290:
### Pre-flight checklist

- [x] I have searched [existing issues](https://github.com/mauriceboe/TREK/issues) and this bug has not been reported yet
- [x] I am running the latest available version of TREK

### TREK version

3.0.0.pre-47

### Describe the bug

half of images is missing in timeline

### Steps to reproduce

go to gallry
go add from synology 
scroll timeline

### Expected behavior

all images should be visible when adding photos from synology

### Deployment method

Docker Compose

### Accessing TREK from

Desktop browser

### Relevant logs or error output

```shell

```

jubnl:
see #693

PR #690, opened 2026-04-16
fix: handling shared albums from synology, pagination now correctly calculate page offset, improving performance on loading small albums

tiquis0290:
## Description
adding passphrases for shared assets and albums,

correctly setting offset in search pagination

changing synology asset id to disallow adding one photo more times when cache id was generated using diffrent call

adding album size to call for rethriving assets, synology has speed problems when requested size is bigger than actual size of album

closing #688 and #689

## Type of Change
- [x] Bug fix
-  New feature
- [x] Breaking change (probably will break old photos)
-  Documentation update

## Checklist
-  I have updated documentation if needed

### DIFF
--- a/client/src/api/client.ts
+++ b/client/src/api/client.ts
@@ -331,8 +331,8 @@ export const journeyApi = {
 
   // Photos
   uploadPhotos: (entryId: number, formData: FormData) => apiClient.post(`/journeys/entries/${entryId}/photos`, formData, { headers: { 'Content-Type': undefined as any } }).then(r => r.data),
-  addProviderPhoto: (entryId: number, provider: string, assetId: string, caption?: string) => apiClient.post(`/journeys/entries/${entryId}/provider-photos`, { provider, asset_id: assetId, caption }).then(r => r.data),
-  addProviderPhotos: (entryId: number, provider: string, assetIds: string[], caption?: string) => apiClient.post(`/journeys/entries/${entryId}/provider-photos`, { provider, asset_ids: assetIds, caption }).then(r => r.data),
+  addProviderPhoto: (entryId: number, provider: string, assetId: string, caption?: string, passphrase?: string | null, cacheId?: string | null) => apiClient.post(`/journeys/entries/${entryId}/provider-photos`, { provider, asset_id: assetId, caption, passphrase, cache_id: cacheId }).then(r => r.data),
+  addProviderPhotos: (entryId: number, provider: string, assetIds: string[], caption?: string, passphrase?: string | null, cacheIds?: string[]) => apiClient.post(`/journeys/entries/${entryId}/provider-photos`, { provider, asset_ids: assetIds, caption, passphrase, cache_ids: cacheIds }).then(r => r.data),
   linkPhoto: (entryId: number, photoId: number) => apiClient.post(`/journeys/entries/${entryId}/link-photo`, { photo_id: photoId }).then(r => r.data),
   updatePhoto: (photoId: number, data: Record<string, unknown>) => apiClient.patch(`/journeys/photos/${photoId}`, data).then(r => r.data),
   deletePhoto: (photoId: number) => apiClient.delete(`/journeys/photos/${photoId}`).then(r => r.data),
--- a/client/src/pages/JourneyDetailPage.tsx
+++ b/client/src/pages/JourneyDetailPage.tsx
@@ -1561,7 +1561,7 @@ function ProviderPicker({ provider, userId, entries, trips, existingAssetIds, on
     setPhotos([])
     setHasMore(false)
     try {
-      const res = await fetch(`/api/integrations/memories/${provider}/albums/${albumId}/photos`, { credentials: 'include', signal })
+      const res = await fetch(`/api/integrations/memories/${provider}/albums/${albumId}/photos${provider === 'synologyphotos' ? `?count=${albums.find(a => a.id === albumId)?.assetCount}` : ''}`, { credentials: 'include', signal })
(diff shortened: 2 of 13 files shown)


jubnl:
Thanks for the work on this. You correctly identified both bugs and the core passphrase gap. However there are several correctness and security issues that block merging as-is.

---

**1. Original downloads are silently broken (logic bug)** 

`server/src/services/memories/synologyService.ts`, `streamSynologyAsset`:

The switch statement maps `kind='original'` to `size='xl'`, then the branch condition is:
```ts
const params = size !== 'original'
  ? new URLSearchParams({ api: 'SYNO.Foto.Thumbnail', ... })
  : new URLSearchParams({ api: 'SYNO.Foto.Download', ... });
```
size is always `'sm'`, `'m'`, or `'xl'`, never the literal string `'original'`. The condition `size !== 'original'` is therefore always `true`, making the `SYNO.Foto.Download` branch permanently unreachable dead code. Every request, including `kind='original'`, is served through the thumbnail API at `xl` (1280 px). Full-resolution originals are no longer downloadable. The correct condition is `kind !== 'original'`.

---

**2. Passphrase stored as plaintext (security regression)**

`server/src/db/migrations.ts` migration 104 and `server/src/services/memories/photoResolverService.ts`, `getOrCreateTrekPhoto`:

Migration 104 adds the column as bare `ALTER TABLE trek_photos ADD COLUMN passphrase TEXT`. The INSERT at line 557 and the UPDATE at line 545 both write the raw passphrase value directly. There is no call to `encrypt_api_key` anywhere in the diff. Every other secret in this codebase: `synology_password`, `synology_sid`, `synology_did`, Immich API keys, is encrypted at rest via `encrypt_api_key` before writing to SQLite. Anyone with read access to the database file can extract passphrases and use them directly against the NAS. The passphrase must be encrypted on write and decrypted on read, consistent with the rest of the credential handling.

---

**3. `cache_key` backfill on existing rows never executes (code bug)**

`server/src/services/memories/photoResolverService.ts`, `getOrCreateTrekPhoto`:

```ts
if (normalizedCacheKey && !existing.id) {
  db.prepare('UPDATE trek_photos SET cache_key = ? WHERE id = ?').run(normalizedCacheKey, existing.id);
}
```

`existing` is the result of `SELECT id, passphrase FROM trek_photos WHERE provider = ? AND asset_id = ? AND owner_id = ?`. existing.id is a primary key, always a non-zero integer. `!existing.id` is therefore always false. The backfill that was supposed to populate cache_key on existing rows never runs.

---

**4. Potential null dereference crash in `_normalizeSynologyPhotoInfo`**

`server/src/services/memories/synologyService.ts`, line 612:

```ts
id: String(item.additional?.thumbnail?.cache_key.split('_')[0] || ''),
```

The optional chaining `?.` stops the chain if `thumbnail` is nullish, but `cache_key` is accessed without `?.` on the result. If `thumbnail` is present as an object but `cache_key` is `null` or `undefined`, which Synology returns for items whose thumbnail has not been generated yet, this throws `TypeError: Cannot read properties of null (reading 'split')` and crashes the entire response for search and album photo listing. This affects all users, not just shared-album scenarios. The original code handled this safely with `item.additional?.thumbnail?.cache_key || ''`.

---

**5. Breaking data migration with no recovery path**

`server/src/db/migrations.ts`, migration 104:

```sql
UPDATE trek_photos
SET cache_key = asset_id,
  asset_id = CASE
    WHEN provider = 'synologyphotos' AND asset_id LIKE '%_%'
    THEN substr(asset_id, 1, instr(asset_id, '_') - 1) 
    ELSE asset_id
  END
WHERE provider = 'synologyphotos' AND asset_id IS NOT NULL
```

This rewrites the `asset_id` of every existing `synologyphotos` row in place. The PR description itself acknowledges: "Breaking change (probably will break old photos)". Any installation with Synology photos already added to trips will see those photos fail to load or disappear after the migration runs, with no automated recovery. A non-breaking approach is to add `cache_key` and `passphrase` as nullable columns without touching existing `asset_id` values.

---

**7. No new tests**

The only changes to `server/tests/integration/memories-synology.test.ts` are six fixture-value substitutions on existing tests replacing `'101_cachekey'` with `'101'` to match the new data model. There are zero new test cases covering shared album listing, passphrase flow through picker and sync, or pagination. Given that this PR modifies the core asset ID scheme, the stream path, and the migration, test coverage for the new behaviour is required.

---

**8. Input validation removed with no substitute **

`server/src/routes/memories/synology.ts`, asset stream route:

The following block is deleted:

```ts
const VALID_SIZES = ['sm', 'm', 'xl'] as const;
const rawSize = String(req.query.size ?? 'sm');
const size = VALID_SIZES.includes(rawSize as any) ? rawSize : 'sm';

if (kind !== 'thumbnail' && kind !== 'original') {
  return handleServiceResult(res, fail('Invalid asset kind', 400));
}
```

With nothing replacing it. Invalid `kind` values now silently fall through to the `default: size = 'sm'` branch in the switch, serving a thumbnail instead of returning a 400.

For all of these reasons, I'll be closing down the PR. I'm working on a fix for the 2 issues at the moment.

tiquis0290:
1. only solution to heic problem same as with immich, didnt realized it made it to this pr,(i would than left a comment about it), intentionaly left it there accesible by change in switch, when there will be proper error handling for loading images in lightbox
2. definitly security regresion should have been ecrypted
3. yeah missed that
4. bug on paper, but yeah missing question mark should been there
5.  if 3 would run than this will worked
6. missing
7. ok
8. removed for fallback didn't though of it as problem, size was never used therefore i removed it in favor of kind

PR #840, opened 2026-04-23
fix(files): attachment double open

### LINKED ISSUE #839: [BUG] Attachment double open
### Pre-flight checklist

### TREK version

3.0.3

### Describe the bug

Clicking the "open" button on an attachment (reservation file, place file, FileManager preview portal, Collab note attachment) opens the file in a new tab and simultaneously triggers the same file in the current tab.

### Steps to reproduce

1. Open a trip.
2. Open a reservation that has an attached PDF or image (or use Files panel → preview → "Open in tab", or a Collab note with a PDF/image attachment).
3. Click the "open in new tab" / file row.

### Expected behavior

The file opens once, in a new browser tab.

### Deployment method

Docker Compose

### Host OS

_No response_

### Accessing TREK from

Desktop browser

### Browser (if applicable)

_No response_

```shell

```

### Screenshots

_No response_

### Additional context

_No response_

Xre0uS:
## Description
`openFile()` in `client/src/utils/fileDownload.ts` opened every safe-inline attachment (PDFs, raster images) **twice**: once in a new tab via `window.open`, and once in the current tab via the popup-blocked download fallback.

`window.open(url, '_blank', 'noreferrer')` always returns `null` because `noreferrer` implies `noopener`, which severs the returned `WindowProxy`. The existing code treated that `null` as "popup blocked" and called `triggerAnchorDownload(blobUrl, filename)`, which clicked an `<a download>` in the current tab on top of the (successfully) opened new tab.

This affected every attachment open path in the app:
- Trip Files panel (`FileManager` lightbox header, PDF preview portal, PDF fallback link)
- Reservation file lists (`ReservationsPanel`, `ReservationModal`)
- Place file lists (`PlaceInspector`)
- Collab note attachments (`CollabNotes` → `FilePreviewPortal`)

### Fix
Replace the `window.open` + null-fallback with a synthetic `<a target="_blank" rel="noopener noreferrer">` click. The browser opens the new tab through normal link handling — no `null`-return ambiguity, no second action in the current tab. PWA standalone behavior is preserved (the fetched blob is same-origin to the PWA WebView, so no system-browser handoff and the session cookie still works). Unsafe MIME types (`text/html`, `image/svg+xml`, …) keep the existing forced-download path.

Public API (`openFile(url, filename?): Promise<void>`) and behavioural contract (async, throws on `!resp.ok`, downloads unsafe MIMEs, opens safe MIMEs in a new tab) are unchanged, so no consumer needs updating.

### Changes
- `client/src/utils/fileDownload.ts` — rewrote the inline-open branch of `openFile()` to use `<a target="_blank" rel="noopener noreferrer">.click()`. Updated doc comment to explain the `window.open` + `noreferrer` `null`-return pitfall.
- `client/tests/unit/utils/fileDownload.test.ts` — updated the existing inline-open test to assert the new anchor-based path (target/rel/no `download` attr) and that `window.open` is **not** called. Added a regression test (`'does not trigger a second in-page action for safe inline types (regression: no double-open)'`) that asserts exactly one anchor click for safe MIME opens. Unsafe-MIME (HTML, SVG) tests unchanged in intent, tightened to `toHaveBeenCalledTimes(1)`.

Closes https://github.com/mauriceboe/TREK/issues/839

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Checklist

### DIFF
--- a/client/src/utils/fileDownload.ts
+++ b/client/src/utils/fileDownload.ts
@@ -32,6 +32,13 @@ function triggerAnchorDownload(blobUrl: string, filename?: string): void {
   setTimeout(() => { URL.revokeObjectURL(blobUrl); a.remove() }, 100)
 }
 
+// navigator.standalone is true only on iOS when running as an
+// add-to-home-screen PWA. In that context, target="_blank" hands off to
+// Safari, which cannot access blob URLs sandboxed to the WebView.
+function isIosStandalone(): boolean {
+  return (navigator as any).standalone === true
+}
+
 /**
  * Fetches a protected file using cookie auth (credentials: include) and
  * triggers a browser download. Works inside PWA standalone mode because the
@@ -56,7 +63,13 @@ export async function downloadFile(url: string, filename?: string): Promise<void
  * (including text/html and image/svg+xml which can execute script) are forced
  * to download so that an uploaded file cannot run code in the TREK origin.
  *
- * Falls back to a download trigger if the popup is blocked.
+ * Uses a synthetic <a target="_blank" rel="noopener noreferrer"> click rather
+ * than window.open(). window.open() called with the "noreferrer"/"noopener"
+ * window feature returns null per spec, which previously made the popup-block
+ * fallback trigger a download in the *current* tab on top of the new-tab open
+ * — i.e. the file opened twice. The anchor approach avoids that ambiguity:
+ * the new tab is opened by the browser's normal link-handling path, and no
+ * spurious in-page download is triggered.
  */
 export async function openFile(url: string, filename?: string): Promise<void> {
   assertRelativeUrl(url)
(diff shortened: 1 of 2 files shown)


client/src/utils/fileDownload.ts
jubnl, line 1: Add a helper after `triggerAnchorDownload`:

```ts
// navigator.standalone is true only on iOS when running as an
// add-to-home-screen PWA. In that context, target="_blank" hands off to
// Safari, which cannot access blob URLs sandboxed to the WebView.
function isIosStandalone(): boolean {
  return (navigator as any).standalone === true;
}
```

Then in openFile, add an early-return before the anchor block:
```ts
// iOS PWA: target="_blank" would open Safari, which can't access the blob
if (isIosStandalone()) {
  triggerAnchorDownload(blobUrl, filename);
  return;
}
```

jubnl (changes requested):
The anchor approach correctly fixes the double-open bug, but introduces a regression on iOS PWA standalone.

On iOS, when TREK is added to the home screen, <a target="_blank"> hands off to Safari instead of opening a new window within the WebView. Safari cannot access blob URLs that were created inside the PWA WebView (they are sandboxed to it), so the file silently fails to open. This is a real use case for us.

Happy to approve these changes once this has been fixed!

Xre0uS:
Thanks for pointing that out! Added the requested chages

jubnl:
Thank you very much for your PR!

PR #898, opened 2026-04-26
(fix) multiply item weight_grams by quantity in bag total weight calc…

gzor:
## Description
Currently the quantity does not effect the total bag weight.
The data is already there, I would just need to change the calculation of the weight of the bag in the frontend.

Use case: User Marco adds lunches with a weight estimation per lunch,  and can adjust the quantity for different trekking trips depending on how long the trek goes.

discord link:
https://discord.com/channels/1488298068427411591/1496176614801670244 

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Checklist

I tested it locally (manual).

I tried to run the test, but they are failing on my machine... i am not sure if i did something wrong or if the tests in general are not up to date. It did not give me the impressions, that the failing tests are associated with my very small change though, therfore i think this is a problem with my local dev env.


gzor:
updated the source branch, and resolved the merge conflicts

mauriceboe:
Thanks @gzor — nice catch, and you're right that the data was already there. 🙌

I pushed a small follow-up onto your branch: pulled the calculation into an `itemWeight()` helper (unit weight × quantity, defaulting to 0 g / qty 1) and applied it to **all** the weight sums — bag totals + max, the unassigned weight, and the grand total, in both the sidebar and the mobile bag modal. A few of those still ignored quantity, so without this the same screen would show inconsistent figures. Added a couple of unit tests for the helper too.

Appreciate the contribution!

PR #1035, opened 2026-05-20
Feat/ldap auth

AlexBander:
## Description
ldap(s) with distinction between admin and user role by group membership

    Add ldapService.ts with bind/search/group-check logic
    Add ldapLoginUser() async wrapper in authService.ts
    Fall back to local login if user not found in LDAP
    Support LDAP_ALLOWED_GROUP for access control
    Support LDAP_ADMIN_GROUP for role mapping
    Support LDAP_TLS_CA for custom CA certificates
    ldapts added as dependency

ENV vars:
LDAP_URL, LDAP_BIND_DN, LDAP_BIND_PW, LDAP_BASE, LDAP_FILTER, LDAP_ADMIN_GROUP, LDAP_ALLOWED_GROUP, LDAP_TLS_CA

Example:
```
# LDAP Authentication (optional)
# If LDAP_URL is set, LDAP authentication is enabled.
# Users not found in LDAP fall back to local authentication.

LDAP_URL=ldaps://ipa.example.com:636
LDAP_BIND_DN=uid=trek-bind,cn=users,cn=accounts,dc=example,dc=com
LDAP_BIND_PW=secret
LDAP_BASE=cn=users,cn=accounts,dc=example,dc=com
LDAP_FILTER=(uid={{username}})
LDAP_ADMIN_GROUP=cn=trek-admins,cn=groups,cn=accounts,dc=example,dc=com
LDAP_ALLOWED_GROUP=cn=trek-users,cn=groups,cn=accounts,dc=example,dc=com
LDAP_TLS_CA=/etc/ssl/certs/ca.crt
```

no discord

## Type of Change
- [ ] Bug fix
- [X] New feature
- [ ] Breaking change
- [ ] Documentation update

## Checklist


mauriceboe:
Thanks a lot for putting this together, @AlexBander — LDAP/LDAPS with group-based role mapping is a genuinely useful addition for self-hosters, and the overall shape (optional, env-gated, fall back to local login, auto-provision, group→admin) is the right one.

I'm going to close this PR, because it can't be merged in its current form:

- The branch was cut from an older `dev`, before the backend moved to NestJS. The LDAP login is wired into `server/src/routes/auth.ts` / `server/src/app.ts`, which no longer exist — login now runs through the Nest `auth-public.controller.ts` → `loginUser`, so `ldapLoginUser` would never actually be called.
- Most of the diff (HEIC conversion, the Maps referer header, the mapbox/PlaceInspector tweaks, the MCP oauth fix, etc.) has since landed on `dev` independently, which is what's causing the conflicts. Only ~4 files are actually LDAP.
- i18n moved from `client/src/i18n/translations/*` into `shared/src/i18n/**` (now behind a strict key-parity CI gate), so those edits no longer apply.

A fresh, LDAP-only PR rebased on current `dev` would be very welcome. A few things to carry over:
- Invoke the LDAP flow from the Nest `AuthService` / `AuthPublicController.login`, not the old Express route.
- `ldapService.ts` reads `process.config.allowGroup` — that's a Node internal, not your config object, so the `LDAP_ALLOWED_GROUP` check never runs (should be `config.allowGroup`).
- Escape the username before building the search filter — `(uid={{username}})` with a raw username is an LDAP-injection vector (use ldapts filter objects or RFC-4515 escaping).
- Add `ldapts` to `server/package-lock.json` (CI runs `npm ci`), and a few tests for the bind / group / role paths.
- Put any new strings in `shared/src/i18n`.

Happy to review the next version. Thanks again!

PR #1086, opened 2026-05-29
Fix inflated trip day_count after dateless → shorter dated updates

### LINKED ISSUE #1083: [BUG] Trip day_count remains inflated after setting shorter date range (dateless days retained)
## Summary
When a trip that has dateless placeholder days (e.g., 7) is later updated to a shorter explicit date range (e.g., 2 days), the trip can still report `day_count = 7`.

This appears to be caused by backend day generation logic retaining extra dateless day rows (`date = null`) after dated days are reassigned.

## Expected behavior
If a trip has explicit `start_date` and `end_date`, the effective day set should match that range (or a clearly defined policy), and `day_count` should not include stale dateless placeholders.

## Actual behavior
- Trip with explicit 2-day range can still have 7 rows in `days` table.
- `day_count` is computed as row count, so it remains 7.
- UI can look inconsistent because some components compute duration from dates while others show `day_count`.

## Suspected root cause
In `server/src/services/tripService.ts`:
- `TRIP_SELECT` computes `day_count` as `COUNT(*) FROM days`.
- `generateDays(...)` (dated path) reuses/renumbers days but can retain unused dateless rows.
- `updateTrip(...)` triggers this path on date/day changes.

## Repro steps
1. Create a trip with no dates and `day_count = 7`.
2. Update trip to `start_date=YYYY-MM-DD`, `end_date=YYYY-MM-DD+1` (2 days).
3. Fetch trip summary/list.
4. Observe only 2 dated days expected, but `day_count` may still be 7 due to extra dateless rows.

### Option A (recommended): normalize days for dated trips
When both `start_date` and `end_date` are set in `generateDays(...)`:
- Keep/remap only the required N days for target dates.
- Remove leftover dateless rows beyond N (at least those with no content).
- Renumber remaining rows consistently.

### Option B: keep extras but exclude from reported `day_count`
Adjust `TRIP_SELECT` to count only relevant days (e.g., dated rows for dated trips). This avoids inflated UI counts but still leaves stale rows in DB.

### Option C: explicit policy flag
Introduce a clear policy/config about preserving dateless days when switching modes, and enforce it consistently in backend + UI.

## Regression tests requested
Please add tests for transitions:
1. `dateless(7) -> dated(2)` should not report 7 unless explicitly intended.
2. `dated(7) -> dated(2)` trims correctly.
3. `dated -> dateless(day_count=X)` preserves/cleans according to policy.

## Notes
I reproduced this behavior on a live MCP-backed instance and confirmed that deleting stale day rows immediately corrected `day_count` and list output.

jordanfelle:
## Summary
- normalize dated trips by removing unused empty dateless rows left over after remapping days
- preserve non-empty dateless rows (with assignments/notes/accommodations) to avoid silent data loss
- keep final day numbering compact and collision-safe

## Why
Fixes issue #1083 where trip day_count remained inflated after switching a trip from dateless placeholders to a shorter explicit date range.

## Tests
- added unit regression: dateless(7) -> dated(2) removes stale placeholders
- added unit coverage: dated -> dateless(day_count=X) trims trailing empty days
- added integration regression: API update from dateless(7) to dated(2) returns day_count=2 and stores only 2 day rows
- ran focused tests in server: 74 passed

Closes #1083

### DIFF
--- a/server/src/services/tripService.ts
+++ b/server/src/services/tripService.ts
@@ -125,13 +125,42 @@ export function generateDays(tripId: number | bigint | string, startDate: string
     del.run(dated[i].id);
   }
 
-  // Any remaining unused dateless days: keep as dateless, just renumber.
-  // Base must be max(targetDates.length, dated.length) to avoid colliding with
-  // positives already assigned by the main loop or the overflow loop above.
-  const maxAssigned = Math.max(targetDates.length, dated.length);
-  for (let i = datelessIdx; i < dateless.length; i++) {
-    setDayNumber.run(maxAssigned + (i - datelessIdx) + 1, dateless[i].id);
+  // Any remaining unused dateless days are stale once explicit dates are set.
+  // To avoid destroying user-entered data, only delete unused dateless days that
+  // are empty (no assignments, notes, or accommodations). Keep non-empty ones.
+  const unusedDatelessIds = dateless.slice(datelessIdx).map(d => d.id);
+  const idsPlaceholders = unusedDatelessIds.map(() => '?').join(', ');
+
+  const datelessWithContent = new Set<number>();
+  if (unusedDatelessIds.length > 0) {
+    const contentRows = db.prepare(
+      `SELECT DISTINCT day_id as id FROM day_assignments WHERE day_id IN (${idsPlaceholders})
+       UNION
+       SELECT DISTINCT day_id as id FROM day_notes WHERE day_id IN (${idsPlaceholders})
+       UNION
+       SELECT DISTINCT start_day_id as id FROM day_accommodations WHERE start_day_id IN (${idsPlaceholders})
+       UNION
+       SELECT DISTINCT end_day_id as id FROM day_accommodations WHERE end_day_id IN (${idsPlaceholders})`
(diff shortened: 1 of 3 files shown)


jordanfelle:
Addressed the Copilot review comments in 9d0fe944:\n\n1) **N+1 query pattern removed**\n- Replaced per-day probing (//) with a set-based approach: one UNION query over the unused dateless IDs, then in-memory  checks for keep/delete decisions.\n\n2) **Inverse retention case covered**\n- Added explicit unit test  to verify an **unused non-empty dateless day** is retained and renumbered when a trip becomes explicitly dated.\n\nValidation run:\n- \n- Result: **74 passed**

jordanfelle:
Addressed the Copilot review comments in commit 9d0fe944.

1. N+1 query pattern removed
- Replaced per-day probing of day_assignments/day_notes/day_accommodations with a set-based path:
  - Build candidate unused dateless IDs once.
  - Run one UNION query (with IN placeholders) to fetch IDs that have content.
  - Use an in-memory Set to decide keep/delete.

2. Inverse retention case covered
- Added explicit unit test TRIP-SVC-019 to verify an unused non-empty dateless day is retained and renumbered when switching to explicit dates.

Validation:
- Ran focused tests in server for unit + integration trip coverage.
- Result: 74 passed.

jordanfelle:
Messed up not making the discord chat first, understand if you need to close

PR #1104, opened 2026-06-04
Fix tsx cache issue

CWSolarDaniel:
Fix tsx cache issue by adding mkdir to Dockerfile

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -24,6 +24,7 @@ COPY --from=client-builder /app/client/dist ./public
 COPY --from=client-builder /app/client/public/fonts ./public/fonts
 
 RUN rm -f package-lock.json && \
+    mkdir -p /tmp/tsx-1000 && \
     mkdir -p /app/data/logs /app/uploads/files /app/uploads/covers /app/uploads/avatars /app/uploads/photos && \
     mkdir -p /app/server && ln -s /app/uploads /app/server/uploads && ln -s /app/data /app/server/data && \
     chown -R node:node /app
--- a/wiki/Development-environment.md
+++ b/wiki/Development-environment.md
@@ -70,47 +70,79 @@ Branch naming conventions:
 
 ## 5. Install Dependencies
 
-Install dependencies for both the client and server:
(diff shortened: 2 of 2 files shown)


jubnl:
What's the issue exactly ? What issue this patch fixes ?

CWSolarDaniel:
> What's the issue exactly ? What issue this patch fixes ?

When the docker image is running, it is causing an issue as follows:
node:fs:1350
  const result = binding.mkdir(
                         ^
Error: ENOENT: no such file or directory, mkdir '/tmp/tsx-1000'
    at Object.mkdirSync (node:fs:1350:26)
    at new FileCache (file:///app/node_modules/tsx/dist/index-7AaEi15b.mjs:13:1080)
    at file:///app/node_modules/tsx/dist/index-7AaEi15b.mjs:13:2208
    at ModuleJob.run (node:internal/modules/esm/module_job:439:25)
    at async node:internal/modules/esm/loader:633:26
    at async asyncRunEntryPointWithESMLoader (node:internal/modules/run_main:96:9) {

  errno: -2,
  code: 'ENOENT',
  syscall: 'mkdir',
  path: '/tmp/tsx-1000'
}
Node.js v24.16.0 

This is on Docker swarm, so that may make a difference.   This patch is to create the directory with the correct permissions so that this issue does not pop up.   I have it running off a build on my fork with just that single line of change.

jubnl:
This is not the right way to fix the issue. Please use something like this instead in your compose file:
```yml
    volumes:
      - type: tmpfs
        target: /tmp
        tmpfs:
          size: 67108864
          mode: 01777
```
Please let me know if that fixed your issue.

CWSolarDaniel:
> This is not the right way to fix the issue. Please use something like this instead in your compose file:
> 
> ```yaml
>     volumes:
>       - type: tmpfs
>         target: /tmp
>         tmpfs:
>           size: 67108864
>           mode: 01777
> ```
> 
> Please let me know if that fixed your issue.

Thank you, I will test, sorry for causing you a bit more work.

PR #1106, opened 2026-06-04
feat(costs): rework Budget into Costs — Splitwise-style, multi-currency, mobile

mauriceboe:
Reworks the Budget addon into **Costs** — a Tricount/Splitwise-style cost tracker — and switches the app font.

## Costs
- **Rename** Budget → "Costs" (UI only; internal addon id, tables and MCP tools unchanged → no breaking changes).
- **Multiple payers** per expense (who actually paid how much) + **equal split** across chosen members.
- **Settle-up**: minimal transfers, persisted history with undo, shared across trip members.
- **12 fixed categories** (users no longer create their own).
- **Per-expense currency** with **live FX conversion** to a user-set **display currency** (Settings → Display). Amounts are stored in their original currency and converted live; closes #551.
- **Locale-correct money formatting** (`Intl.NumberFormat`: symbol position, separators, decimals per currency/locale).
- **Desktop + dedicated mobile layout**; the bottom-nav "+" adds an expense on the Costs tab.
- **Migration** backfills existing budget items into the new model (single payer → one payer, split members, currency) and assigns legacy categories to "Other".

## Font
- App font switched to **self-hosted Poppins** (with **Geist** for secondary subtext), removing the Google Fonts CDN dependency.

## Also
- Tightens authorization on the journey share-link read endpoint (`canAccessJourney`, matching the create/delete routes and the MCP tool).

Verified: client + server `tsc` clean, i18n parity green (20 locales), budget + journey test suites green.


upendedshark:
This is a great feature, but how do i remove cost splitting on a trip? most of my trips are with my own family where my partner and I are not splitting costs

PR #1201, opened 2026-06-16
feat: switch currency widget to Frankfurter v2 API

melanki:
## Summary
- Replaces `exchangerate-api.com` with [`frankfurter.dev`](https://frankfurter.dev) (v2) as the rate source for the dashboard CurrencyWidget.
- Updates the server CSP `connect-src` allowlist accordingly, plus the MSW intercept in `DashboardPage.test.tsx` and the wiki entry in `Dashboard-Widgets.md`.
- Same `data.rates[to]` field is read, so the widget behavior is unchanged. The new endpoint shape is `GET /v2/rates?base={from}&quotes={to}` returning `{ base, date, rates: { [code]: number } }`.

## Why
Frankfurter is open-source, aggregates data from central banks, and supports historical lookups by date. That historical capability is the reason for the swap — it will be used in a follow-up PR to convert prior-dated expenses for multi-currency budgets. This PR only does the API swap; no UI change.

## Test plan
- [x] `npx vitest run src/pages/DashboardPage.test.tsx` — 32/32 pass with the updated MSW intercept.
- [ ] Manually try the CurrencyWidget on the dashboard: change source/target currency, click refresh — rate should update.
- [ ] Confirm the CSP doesn't block the request in the browser console.

### DIFF
--- a/client/src/components/Dashboard/CurrencyWidget.tsx
+++ b/client/src/components/Dashboard/CurrencyWidget.tsx
@@ -31,7 +31,7 @@ export default function CurrencyWidget() {
     if (from === to) { setRate(1); return }
     setLoading(true)
     try {
-      const resp = await fetch(`https://api.exchangerate-api.com/v4/latest/${from}`)
+      const resp = await fetch(`https://api.frankfurter.dev/v2/rates?base=${from}&quotes=${to}`)
       const data = await resp.json()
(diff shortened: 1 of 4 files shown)


jubnl:
v3.1.0 now use https://frankfurter.dev/ to get the exchange rates

PR #1301, opened 2026-06-24
feat(extract): extract data using LLM

jubnl:
## Description

Adds an optional LLM fallback to the booking import. When kitinerary can't read a confirmation (no barcode, an unusual layout, plain text), the extracted text goes to a configured model and the result is mapped through the same pipeline kitinerary uses, so nothing downstream changes.

How it behaves:
- AI Parsing addon off: kitinerary only, exactly like before.
- Addon on: if kitinerary finds nothing, the model runs automatically. No extra confirmation step and no per-user toggle.

Providers are local (any OpenAI-compatible server such as Ollama), OpenAI and Anthropic, configured under Admin > Addons. A NuExtract model gets its own prompt format when it's selected: these models are trained to fill a JSON template rather than follow chat instructions, so the template is passed inline and the temperature pinned to 0, otherwise they just echo the schema back. On a CPU-only host NuExtract 2.0 2B is the recommended default since it's small, fast and MIT licensed.

What the import fills:
- Flights, trains, buses, boats, rental cars, hotels, restaurants and events.
- From/To endpoints for transport, geocoded with Nominatim when neither the model nor kitinerary give coordinates, so routes and map pins show up.
- Booking codes, pickup/return and check-in/out times, seat, class and platform.
- A linked cost in the matching category when a confirmation carries a price. Prices are read as text and parsed locally so a German "1.580,22 €" doesn't end up as 1.49772.

Instead of writing everything straight to the trip, the import opens the normal edit modal pre-filled for each parsed booking. You check it, fix anything the model got wrong (a guessed year, a wrong address) and save. Hotels get an editable address field, and on save an existing place is matched by name or a new one is created from the address you confirmed.

Notes:
- The extracted text is capped so a long terms-and-conditions tail doesn't overflow a small model's context window or drag out CPU inference.
- A scanned, image-only PDF has no text layer to read. Anthropic handles those natively; the local/OpenAI path needs a text layer.

## Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update

## Checklist


mauriceboe:
not ready yet, working on it again tomorrow!

PR #879, opened 2026-04-24
Feat/calendar sync

tiquis0290:
## Description
Implementing pernament ics calendar link accesible by share token, for adding calendars to app, which then can donwload changes

## Related Issue or Discussion
feat Add calendar sync feature using iCal link
https://discord.com/channels/1488298068427411591/1492952040484114432

## Type of Change
-  Bug fix
- [x] New feature
-  Breaking change
- Documentation update

## Checklist
- [x] I have read the [Contributing Guidelines](https://github.com/mauriceboe/TREK/wiki/Contributing)
- [x] My branch is [up to date with `dev`](https://github.com/mauriceboe/TREK/wiki/Development-environment#3-keep-your-fork-up-to-date)
- [x] This PR targets the `dev` branch, not `main`
- [x] I have tested my changes locally
- [x] I have added/updated tests that prove my fix is effective or that my feature works
- [x] I have updated documentation if needed

PR #1307, opened 2026-06-25
feat(feeds): subscribable ICS calendar feeds for trips

michael-bohr:
## Description

Adds TripIt-style **subscribable** calendar feeds alongside the existing one-time
.ics download. A trip (or all of a user's trips) gets a secret, revocable feed URL
that Google/Apple/Outlook poll to stay in sync.

- Public token-based read endpoints `GET /api/feed/trip/:token.ics` and `/api/feed/user/:token.ics`
  (no login — the secret token is the credential), reusing `exportICS()` and adding
  REFRESH-INTERVAL / X-PUBLISHED-TTL hints.
- JWT-guarded endpoints to generate (lazy) and regenerate/revoke per-trip and
  per-user feed tokens; stored in two nullable `feed_token` columns.
- All-trips feed excludes archived trips and trips ended >90 days ago.
- UI: ICS button → Download/Subscribe menu; modal with one-click "Add to Google
  Calendar" and a `webcal://` link for Apple/Outlook, plus copy fallbacks.
- Feed base URL from the existing `APP_URL` env var. Purely additive, no breaking changes.

Picks up the approved-but-stale #879 (closed in cleanup; root cause was the
Express→NestJS migration). Reimplemented cleanly on the current NestJS architecture.

Addresses discussion: https://discord.com/channels/1488298068427411591/1492952040484114432

## Type of Change
- [x] New feature

## Test plan
- `server/tests/e2e/feeds.e2e.test.ts` (10 tests, modeled on `share.e2e.test.ts`):
  token generate lazy + idempotent, regenerate invalidates old, 401/404 auth+access,
  public feed content-type + hint injection, unknown-token 404, all-trips
  archived/>90-day exclusion.
- Manual: subscribed `webcal://` in Apple Calendar and the `render?cid=` deep link
  in Google Calendar against a live instance — events appear and reflect edits.


alghamdhyl-arch (approved):

mauriceboe:
Thanks a lot for this, @michael-bohr — really nice work. 🙏 This was the stronger of two parallel takes on calendar feeds (the other was #1357): reusing `exportICS()` as the single source of truth, and shipping the per-trip + all-trips subscribe UX (webcal / Add-to-Google), is exactly the right shape.

Heads up — I pushed a hardening commit on top of your branch (so it's been force-rebased onto the latest `dev`) to cover a few things before merge:

- **Absolute feed URLs.** They now fall back to the request host when `APP_URL` isn't set, so the webcal / Google links work on a default `docker run` — otherwise `feedUrl()` returned a relative path and the `webcal://` rewrite was a no-op.
- **A real off switch.** `POST` enables, `PUT` rotates, `DELETE` clears the token (`feed_token = NULL`), and the dialog no longer mints a token just from being opened — opening it only reads.
- **Smaller fixes:** octet-safe RFC 5545 line folding in `exportICS` (borrowed from #1357), VEVENT extraction by line scan instead of a regex that user text could truncate, URL-encoded Google `cid`, `feed_token` mirrored into `schema.ts`, and the duplicated all-trips modal folded into the shared `IcsSubscribeModal`. Tests updated to match (e2e for rotate/disable/host-fallback).

All green and good to merge. Thanks again for the clean implementation — this is a great addition. 🚀

PR #1332, opened 2026-06-27
fix(maps): increase Overpass timeout and add instance-specific User-Agent

### LINKED ISSUE #1309: [BUG] Points of Interest Search Fails
### Pre-flight checklist

### TREK version

3.1.2

### Describe the bug

When using the POI search in the map view the API (/api/maps/pois) always returns 502.

### Steps to reproduce

1. Open a trip plan
2. Click on any of the POI search buttons

### Expected behavior

POIs to be returned

### Deployment method

Kubernetes / Helm

### Host OS

talos

### Accessing TREK from

Desktop browser

### Browser (if applicable)

Safari 26.5, Chromium 148.0.7778.167

```shell
error in debug logs for 2x locations:


/api/maps/pois 502 10504ms ip=10.244.1.0 query={"category":"restaurant","south":"-41.2007080775279","west":"146.3180807743064","north":"-41.16592608250532","east":"146.3752986241381"}



GET /api/maps/pois 502 10505ms ip=10.244.1.0 query={"category":"bar","south":"48.85903221273091","west":"2.3365771816587824","north":"48.86936011436177","east":"2.3588906176328237"}
```

### Screenshots

Image

### Additional context

_No response_

Hardik-369:
Fixes #1309

The POI search endpoint (/api/maps/pois) returned 502 errors for two reasons:

1. **Timeout mismatch**: OVERPASS_TIMEOUT_MS (12s) was shorter than actual mirror response times. The reporter confirmed kumi.systems takes ~25.7s to first byte. Increased to 25s to match the [timeout:20] query timeout.

2. **Non-unique User-Agent**: The static UA string was identical across all deployments, making rate-limiting more likely. The overpass-api.de documentation recommends a unique identifier per application/instance. The new userAgent() function appends the instance's APP_URL so each deployment identifies itself uniquely.

Tested by the reporter: overpass-api.de responds in ~0.96s with a proper User-Agent, and the kumi.systems/private.coffee mirrors were found to need >25s for first byte.


mauriceboe:
Thanks @Hardik-369 — the instance-specific User-Agent is a good catch and I've kept it. I've reworked the rest on top of your two commits and retargeted this at the `v3-1-3` release branch:

- **Unique User-Agent** kept (`buildUserAgent`), applied to every Nominatim/Overpass/Wikimedia call.
- **`OVERPASS_URL`** (new): operators behind locked-down egress — like the reporter's K8s cluster — can point the explore search at an internal/self-hosted Overpass instance instead of the public mirrors. This is the durable fix for the unreachable-mirrors case.
- **Timeout**: I returned the default to 12s (raising it doesn't change the all-mirrors-down outcome, it just makes users wait longer for the same 502) but made it tunable via **`OVERPASS_TIMEOUT_MS`** for slow self-hosted instances. Non-positive/invalid values fall back to the default rather than aborting every request at a 0 ms cap.
- **Clearer failure**: each endpoint's failure reason is now logged before the 502, so a blocked-egress/firewall situation is diagnosable instead of a bare "Overpass request failed".
- Added unit tests for the User-Agent, endpoint and timeout resolution plus the all-mirrors-down path, and documented the two new env vars in `.env.example`, the wiki and the Helm chart.

PR #1476, opened 2026-07-06
fix(oauth): don't revoke active MCP sessions on a normal token refresh

### LINKED ISSUE #1475: [BUG] MCP /mcp returns 404 after OAuth access-token refresh (unconditional session revoke on token rotation)

### Pre-flight checklist

### TREK version

3.1.4

### Describe the bug

After the MCP OAuth 2.1 access token refreshes (which happens automatically about once an hour, since access tokens are issued with a 1-hour TTL), every subsequent request from the still-connected MCP client to `/mcp` starts returning `404 { "error": "Session not found" }`. The MCP integration is otherwise completely broken until the client is manually disconnected and reconnected in the AI tool (Claude Code), which forces a brand-new session.

Root cause: `refreshTokens()` in `server/src/services/oauthService.ts` unconditionally calls `revokeUserSessionsForClient(row.user_id, clientId)` on **every successful** refresh-token grant, not only when refresh-token replay/theft is detected:

```ts
// server/src/services/oauthService.ts (refreshTokens)
// Revoke old pair immediately (rotation) and issue new pair linked to old row
db.prepare('UPDATE oauth_tokens SET revoked_at = CURRENT_TIMESTAMP WHERE id = ?').run(row.id);

// Terminate active MCP sessions for the old token's client so client must re-authenticate
revokeUserSessionsForClient(row.user_id, clientId);
```

This deletes the live, in-memory MCP session (`server/src/mcp/sessionManager.ts`) — including the open SSE transport the MCP client is actively using — purely because the access token rotated. Since the MCP `mcp-session-id` is independent of the bearer token, and the session lookup already re-validates `session.userId` and `session.clientId` against the new token on every request (`server/src/mcp/index.ts`, `mcpHandler`), there's no security reason to kill the session here; a normal, non-replayed refresh is not an attack signal.

There's a separate, correctly-scoped code path a few lines above that revokes sessions specifically on **refresh-token replay detection** (an actual theft indicator) — that one makes sense. The unconditional revoke on the normal rotation path looks like an over-broad copy of that logic, introduced in the OAuth/MCP security-hardening pass (commit `7c0a0d5f`, finding C2 "wire token/session revocation into active MCP session lifecycle").

The net effect: any long-running MCP client (Claude Code, Claude Desktop, etc.) using OAuth 2.1 breaks itself roughly every hour, which defeats the purpose of the configurable session TTL / SSE keep-alive (`MCP_SESSION_TTL`, `MCP_SSE_KEEPALIVE`) that exists specifically to keep long MCP sessions alive.

### Steps to reproduce

1. Register an OAuth 2.1 client in TREK (Settings → Integrations → MCP → OAuth Clients) and connect an MCP client (e.g. Claude Code) using OAuth.
2. Use the MCP connection normally, and leave it connected/idle for longer than the access-token TTL (default 1 hour).
3. When the MCP client's OAuth layer silently refreshes its access token in the background, watch the TREK server logs.
4. Observe `[INFO] ... oauth.token.refresh ...` immediately followed by repeated `[WARN] ... POST /mcp 404 ...` / `[WARN] ... GET /mcp 404 ...` for every request the MCP client makes afterward.
5. The MCP tool stops working until you manually disconnect/reconnect the MCP server in the client.

### Expected behavior

A normal (non-replayed) OAuth refresh-token grant should **not** terminate the caller's own active MCP session. The MCP connection should continue working uninterrupted across access-token refreshes, for as long as the session is otherwise active (per `MCP_SESSION_TTL` / SSE keep-alive). Session revocation on refresh should be limited to the replay-detection branch (token reuse / theft), not every legitimate rotation.

### Deployment method

Docker Compose

### Host OS

Synology DSM

### Accessing TREK from

Desktop browser

```
07/07/2026 12:54:37 AM [MCP] Session sweep: cleaned 0, active 2
07/07/2026 12:55:37 AM [MCP] Session sweep: cleaned 0, active 2
07/07/2026 12:56:23 AM [INFO] 2026-07-07T00:56:23 zzzzzzzzz@gmail.com oauth.token.refresh ip=192.168.xx.yyy
07/07/2026 12:56:23 AM [WARN] 2026-07-07T00:56:23 POST /mcp 404 1ms ip=192.168.xx.yyy
07/07/2026 12:56:24 AM [WARN] 2026-07-07T00:56:24 GET /mcp 404 1ms ip=192.168.xx.yyy
07/07/2026 12:56:24 AM [WARN] 2026-07-07T00:56:24 GET /mcp 404 1ms ip=192.168.xx.yyy
07/07/2026 12:56:26 AM [WARN] 2026-07-07T00:56:26 GET /mcp 404 1ms ip=192.168.xx.yyy
07/07/2026 12:56:26 AM [WARN] 2026-07-07T00:56:26 GET /mcp 404 1ms ip=192.168.xx.yyy
07/07/2026 12:59:59 AM [WARN] 2026-07-07T00:59:59 POST /mcp 404 1ms ip=192.168.xx.yyy
```

### Additional context

- TREK is running behind a Traefik reverse proxy, but this is not a proxy/timeout issue — the 404 responses return in ~1ms, i.e. the app itself is answering, not a dropped/timed-out connection. Traefik is incidental here.
- The only workaround currently is to manually reconnect the MCP server in the client (e.g. `/mcp reconnect` in Claude Code), which forces creation of a brand-new session (no `mcp-session-id` header sent), bypassing the dead session entirely.
- Suggested fix: in `refreshTokens()` (`server/src/services/oauthService.ts`), remove (or scope down) the `revokeUserSessionsForClient(row.user_id, clientId)` call on the normal-rotation path, keeping it only in the replay-detected branch. The per-request session validation in `mcpHandler` already re-checks `userId`/`clientId` against the current token on every call, so a legitimate session doesn't need to be pre-emptively killed on refresh.

daften:
## Summary
- `refreshTokens()` unconditionally called `revokeUserSessionsForClient(...)` on every successful refresh-token grant, not just on replay/theft detection.
- Since MCP access tokens expire after 1 hour, any long-lived MCP client (Claude Code, Claude Desktop, etc.) silently refreshing its token in the background had its live MCP session killed roughly every hour, causing `/mcp` to return `404 { error: "Session not found" }` until the client was manually reconnected.
- Session revocation now happens only in the replay-detection branch (an actual security event); `mcpHandler` already re-validates `session.userId`/`session.clientId` against the current token on every request, so a legitimate rotation doesn't need to pre-emptively kill the session.

Fixes #1475

## Test plan
- [x] Added a unit test asserting a normal (non-replayed) refresh does **not** call `revokeUserSessionsForClient`
- [x] Added a unit test asserting refresh-token replay **still** revokes sessions (unchanged security behavior)
- [x] `npx vitest run tests/unit/services/oauthService.test.ts` — 70/70 passing
- [x] `npx vitest run tests/unit/nest/oauth.service.test.ts tests/unit/nest/oauth.controller.test.ts` — 62/62 passing
- [x] `npx vitest run tests/e2e/oauth.e2e.test.ts` — 6/6 passing
- [x] `npm run typecheck` — clean
- [x] `npx eslint src/services/oauthService.ts tests/unit/services/oauthService.test.ts` — no new warnings


mauriceboe:
Thanks for the fix! This PR targets `dev`, which has diverged too far from the `v3-2-2` release branch to merge cleanly, so I cherry-picked your work onto `v3-2-2` with your authorship preserved (`570d2423`). It will ship in 3.2.2 — closing here since the fix is now in the release branch. Much appreciated!

daften:
HI @mauriceboe Thanks so much! Just wanted to indicate I (mostly 🙈 ) followed the directions in CONTRIBUTING.md, that's why I targeted it at the dev branch. :) The fourth item. I think that's an important one so you can avoid rebase or cherry-picking work in the future. :)

Issue #1366, opened 2026-06-29
[BUG] Only countries with associated lodging are considered as visited


svillar:
### Pre-flight checklist

- [x] I have searched [existing issues](https://github.com/mauriceboe/TREK/issues) and this bug has not been reported yet
- [x] I am running the latest available version of TREK
- [x] I have read the [Troubleshooting guide](https://github.com/mauriceboe/TREK/wiki/Troubleshooting) and my issue is not covered there

### TREK version

3.1.2

### Describe the bug

I have a 1 day trip to the capital of one country which is really close to the border of another country where I'm staying for some days. I've already added transport but interestingly that is not enough to consider the country as visited. IMHO any kind of activity, like transport, plans to visit, etc... should be considered to assess whether a country is visited and not only the lodging

### Steps to reproduce

1. Add some transport booking to a new country which does not have any associated lodging
2. Check the main dashboard

### Expected behavior

The country should be considered as visited

### Deployment method

Docker Compose

### Host OS

Debian stable

### Accessing TREK from

Desktop browser

### Browser (if applicable)

Firefox 140

### Relevant logs or error output

```shell

```

mauriceboe:
Fixed — countries you reach only via a transport booking (no lodging or place there) now count as visited too, on both the dashboard and the Atlas page. Ships in **v3.2.0**.

PR #1488, opened 2026-07-07
fix(atlas): connecting-flight layovers no longer mark the layover country visited

### LINKED ISSUE #1486: [BUG] Connecting-flight layover marks the layover country as "visited" in Atlas

### Pre-flight checklist

### TREK version

3.1.4

### Describe the bug

The Atlas page marks a country as "visited" even when the only reason it was ever geocoded is that a connecting flight touched down there during a layover — the traveler never left the airport.

Root cause: `getStats()`'s "countries reached only by a transport booking" merge (`server/src/services/atlasService.ts`, `#1366`) reads every row in `reservation_endpoints` for the user's trips and geocodes each `(lat, lng)` to a country, to catch cases where someone flew into a country with no lodging/place logged. The query never filtered on the endpoint's `role` column, so a `role: 'stop'` row — the intermediate airport of a multi-leg flight recorded as a single reservation (`from → stop → to`) — is treated exactly like a genuine arrival/departure. Any connecting flight through a given country's airport therefore marks that entire country "visited," regardless of how short the layover was.

### Steps to reproduce

1. Add a flight reservation with a layover (Planner → add Reservation → type Flight → use "+ Add Stop" to insert a connecting airport between origin and destination), e.g. `Brussels → Helsinki (stop) → New York`.
2. Open the Atlas page.
3. Observe that Finland is shown as a visited country, filled in on the map and counted in the country total, even though no place/lodging was ever logged there and the layover was under an hour.

### Expected behavior

A `stop` (layover/connection) endpoint should not count as having "visited" that country. Only a flight's actual origin/destination endpoints (`role: 'from' | 'to'`) should be eligible for the transport-only country merge.

### Deployment method

Docker Compose

### Host OS

Synology DSM

### Accessing TREK from

Desktop browser

### Additional context

Fix: filter the merge query to `e.role IN ('from', 'to')`, excluding `stop` rows.

A related, separate finding from the same investigation: a `car`/driving reservation's destination can also be miscounted if the geocoded waypoint happens to sit on the wrong side of a border shared by two same-named towns (e.g. "Niagara Falls" exists as both a US and a Canadian city) — that one is a geocoding-precision edge case rather than a `role` issue, and is not fixed by this PR.

daften:
## Description

`getStats()`'s `#1366` "transport-only country" merge (`server/src/services/atlasService.ts`) read every `reservation_endpoints` row for a trip and geocoded it to a country, without filtering on the endpoint's `role`. A `role: 'stop'` row — the intermediate airport of a multi-leg flight (`from → stop → to`) — was treated identically to a real arrival/departure, so a connecting flight through e.g. Helsinki marked Finland as visited even on a 50-minute layover with no place ever logged there.

Fixed by filtering the merge to `e.role IN ('from', 'to')`, excluding `stop` rows. Confirmed against a real backup (not just the sample/dev dataset), and re-verified live: toggling the fix off/on while holding the data fixed reliably reproduces/resolves the symptom.

A related but separate finding from the same investigation, **not** fixed here: a `car`/driving reservation's destination can also be miscounted if the geocoded waypoint sits on the wrong side of a border shared by two same-named towns (e.g. "Niagara Falls" exists as both a US and a Canadian city) — that's a geocoding-precision problem, not a `role` issue, called out in the issue for follow-up.

Closes #1486

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Checklist
- [ ] I have updated documentation if needed — *n/a, no wiki page described the buggy behavior as intended*


mauriceboe:
Thanks for the fix! This PR targets `dev`, which has diverged too far from the `v3-2-2` release branch to merge cleanly, so I cherry-picked your work onto `v3-2-2` with your authorship preserved (`e419dcbd`). It will ship in 3.2.2 — closing here since the fix is now in the release branch. Much appreciated!

PR #1489, opened 2026-07-07
fix(atlas): match region names regardless of accent/dash formatting

### LINKED ISSUE #1487: [BUG] Atlas sub-region highlighting fails when cached region name differs by accent/dash from the map data (e.g. Île-de-France)

### Pre-flight checklist

### TREK version

3.1.4

### Describe the bug

On the Atlas page, zooming into a visited country to see its sub-national regions highlighted, a region that was genuinely visited (with logged places) can fail to show as highlighted at all — e.g. Île-de-France (Paris) not highlighting despite multiple logged Paris trips.

Root cause: the region layer (`client/src/pages/atlas/useAtlas.ts`) matches a boundary feature against the visited-regions list first by ISO 3166-2 code, then falls back to comparing region *names* (since the geocoder's cached region code frequently doesn't match the bundled admin-1 boundary's code — a separate, known mismatch). That name fallback was a plain `.toLowerCase()` string comparison. The cached `region_name` (from Nominatim reverse-geocoding, stored in `place_regions`) and the bundled boundary's `name`/`name_en` property don't always agree on accenting or punctuation — e.g. the cache holds `"Ile-de-France"` (no accent) while the bundle has `"Île-de-France"`, or `"Bourgogne – Franche-Comté"` (en dash) vs. the bundle's `"Bourgogne-Franche-Comté"` (hyphen). A plain lowercase comparison treats these as different strings, so the fallback silently fails and the region never highlights.

### Steps to reproduce

1. Log a trip with a place in Paris (or any place whose geocoded region name happens to be cached without accents, e.g. from an older/pre-migration geocode).
2. Open Atlas, zoom into France (zoom level ≥ 6 to load the region layer).
3. Observe Île-de-France is not filled/highlighted as visited, even though the country-level stats correctly count places there.

### Expected behavior

A region should highlight as visited whenever its name matches the cached visited-region name, independent of accenting, diacritics, or hyphen vs. en/em-dash style.

### Deployment method

Docker Compose

### Host OS

Synology DSM

### Accessing TREK from

Desktop browser

### Additional context

Fix: added a `normalizeRegionName()` helper (`client/src/pages/atlas/atlasModel.ts`) that strips diacritics and folds dash variants before comparing, used everywhere the region layer compares names.

A related, deeper mismatch also surfaced in the same investigation but is **not** fixed by this PR: some cached region codes/names come from a Nominatim address level (`ISO3166-2-lvl6`, department-level for France) that doesn't correspond to the bundled boundaries at all (region-level), and some region names are cached in a different language than the bundle uses (e.g. cached "Brittany" vs. bundle's "Bretagne") — those would need a broader fix to the geocoding priority/translation logic in `buildRegionInfo()` (`server/src/services/atlasService.ts`).

daften:
## Description

The Atlas region layer (`client/src/pages/atlas/useAtlas.ts`) falls back to comparing region *names* when the geocoder's cached region code doesn't match the bundled admin-1 boundary code (a separate, pre-existing mismatch). That fallback did a plain `.toLowerCase()` compare, so a cached name like `"Ile-de-France"` (no accent, as reverse-geocoded) never matched the bundle's `"Île-de-France"`, and `"Bourgogne – Franche-Comté"` (en dash) never matched `"Bourgogne-Franche-Comté"` (hyphen) — silently leaving genuinely visited regions unhighlighted.

Added `normalizeRegionName()` (`client/src/pages/atlas/atlasModel.ts`) — strips diacritics, folds dash variants — and used it everywhere region names are compared. Confirmed against a real account's data that Île-de-France highlights correctly after the fix.

A related, deeper mismatch also surfaced in the same investigation but is **not** fixed here: some cached region codes/names come from a Nominatim address level (department-level for France) that doesn't correspond to the bundled boundaries at all (region-level), and some region names are cached in a different language than the bundle uses (e.g. cached "Brittany" vs. bundle's "Bretagne") — called out in the issue for follow-up, since it needs broader work in the geocoding priority/translation logic.

Closes #1487

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Checklist
- [ ] I have updated documentation if needed — *n/a, no wiki page described the buggy behavior as intended*


mauriceboe:
Thanks for the fix! This PR targets `dev`, which has diverged too far from the `v3-2-2` release branch to merge cleanly, so I cherry-picked your work onto `v3-2-2` with your authorship preserved (`e5e3db99`). It will ship in 3.2.2 — closing here since the fix is now in the release branch. Much appreciated!

PR #1519, opened 2026-07-11
fix(pdf): use the trip's actual currency instead of hardcoded EUR

trongbinh15:
## Summary
image

- Trip PDF export (day totals, place price chips, cover total-cost stat) now formats amounts with the trip's/place's actual currency via `formatMoney`, instead of a hardcoded `EUR` suffix.
- `pdf.costLabel` no longer bakes currency text into the label across the 21 locales that had it (Hungarian was already generic).

## Test plan
- [x] `cd client && npx vitest run src/components/PDF/TripPDF.test.ts`
- [x] `cd shared && npm run i18n:parity:strict`
- [x] `cd shared && npm test`
- [x] `npm run lint` (all workspaces)

### DIFF
--- a/client/src/components/PDF/TripPDF.test.ts
+++ b/client/src/components/PDF/TripPDF.test.ts
@@ -286,13 +286,45 @@ describe('downloadTripPDF', () => {
     expect(iframe!.srcdoc).toContain('CONF999')
   })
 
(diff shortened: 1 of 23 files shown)


jubnl:
lgtm, thx!

PR #1551, opened 2026-07-14
feat(mcp): expose public transit planning tools

Uziniii:
## Description

Expose the existing public transit functionality through MCP tools. MCP clients can search for stops and routes, then save a selected journey to a trip.

## Related Issue or Discussion

Discussed in the [`#github-pr` Discord thread](https://discord.com/channels/1488298068427411591/1526341000673493133).

## Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update

## Checklist
- [x] I have read the [Contributing Guidelines](https://github.com/mauriceboe/TREK/wiki/Contributing)
- [x] My branch is [up to date with `dev`](https://github.com/mauriceboe/TREK/wiki/Development-environment#3-keep-your-fork-up-to-date)
- [x] This PR targets the `dev` branch, not `main` *(wiki-only PRs are exempt)*
- [x] I have tested my changes locally
- [x] I have added/updated tests that prove my fix is effective or that my feature works
- [x] I have updated documentation if needed

PR #1558, opened 2026-07-14
feat(mcp): expose public transit planning tools

jubnl:
## Description

Exposes the existing public-transit functionality (#1065) through MCP: `search_transit_stops`, `search_transit_routes` and `create_transit_journey`, so an MCP client can search stops and routes and save a selected journey to a trip day.

This carries @Uziniii's work from #1551 unchanged (their commits are preserved) and adds one fix on top. It supersedes #1551, which is on a fork branch we cannot push to — credit for the feature is entirely theirs.

### The fix on top

`search_transit_routes` validated each itinerary leg's `mode` against `SCHEDULED_TRANSIT_MODES`. That constant is the **request-side** filter whitelist — the modes a caller may *ask* for — not the modes MOTIS can *return*. Per the MOTIS spec, the default `TRANSIT` mode expands to `TRAM,FERRY,AIRPLANE,BUS,COACH,RAIL,ODM,RIDE_SHARING,FUNICULAR,AERIAL_LIFT,OTHER`, and street legs can be `BIKE`/`CAR`/`RENTAL`. An itinerary carrying any of those failed the parse and was discarded by the `flatMap`, so the tool reported fewer routes than exist — or none.

Against the live provider, Trondheim → Ålesund (Norway's GTFS includes air routes):

```
itineraries returned by Transitous: 5
itineraries the MCP tool would silently drop: 3
  dropped: [WALK, BUS, WALK, AIRPLANE, WALK, AIRPLANE, WALK, BUS, WALK, COACH, WALK]
```

The web app shows all 5 — `TransitSearchPanel` treats `mode` as a free string and renders anything non-`WALK` as a transit leg.

So a leg's `mode` is now accepted as any mode token, with the existing "at least one non-`WALK` leg" rule doing the real gating. Everything downstream keeps its `mode !== 'WALK'` semantics, so a journey created over MCP is byte-identical to one created in the app.

Dropping an itinerary is still possible when provider data fails the remaining consistency rules (leg duration vs. its times, adjacent-leg continuity, leg times within the itinerary window) — but a silent drop is indistinguishable from "no routes exist", and the assistant would confidently report the latter. `search_transit_routes` now returns a `dropped` count alongside the results.

Supersedes #1551. Discussed in the [`#github-pr` Discord thread](https://discord.com/channels/1488298068427411591/1526341000673493133). Builds on the transit feature from #1065.

## Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update

## Checklist

---

Two regression tests were added to `server/tests/unit/mcp/tools-transit.test.ts` — an `AIRPLANE`-leg itinerary surviving search and persisting through create, and the `dropped` count being reported. Both fail on #1551 as it stands. Full server suite: 297 files / 5,229 tests passing; typecheck and lint clean.


Uziniii:
thx, it was quick to merge, as a note for you that ive flagged as duplicate and gap validation between frontend and rest api that the mcp does (solve? im not an english native so sorry if its not clear) : 

Other duplications identified:
- Client-side transit reservation construction: TransitSearchPanel.tsx:378-470 duplicates name cleanup, endpoint construction, timezone conversion, and metadata.transit generation now available in transitItineraryService.ts.
- Daily route calculation: useRouteCalculation.ts:49-185 and DayPlanSidebar.tsx:476-549 both reconstruct road segments, transport breaks, and hotel connections.
- Transport type lists: TRANSPORT_TYPES is declared in multiple client files and should be shared.
- Reservation side effects: createReservation → broadcast → notifyBookingChange is repeated across MCP tools, Nest controllers, and plugins.
- Endpoint-chain construction: similar from → stop → to logic exists in AirTrail, KItinerary, MCP transport, and transit mappers.
- Haversine calculation: several client files implement their own version instead of using a shared utility.
- REST/MCP validation gap: MCP applies strict transit-specific validation, while REST currently relies on client-built reservation data and the generic reservation endpoint.
The highest-value remaining duplication is the daily route calculation shared between useRouteCalculation.ts and DayPlanSidebar.tsx.

PR #1563, opened 2026-07-14
fix(atlas): resolve region AND country by coordinates against the bundled polygons

### LINKED ISSUE #1547: [BUG] Atlas region highlighting still fails when Nominatim's geocode level doesn't match the bundled boundary level (e.g. Barcelona, Madrid)

### Pre-flight checklist

### TREK version

3.3.0

### Describe the bug

Follow-up to #1487. That fix (accent/dash-insensitive name matching) resolved the Île-de-France case, but places in Barcelona and Madrid still don't highlight their region on the Atlas map, and it's not an accenting problem this time.

Root cause: Nominatim's reverse-geocode address level (province, autonomous community, borough, …) doesn't always correspond to whatever granularity the bundled geoBoundaries admin-1 data happens to ship for a given country. For Spain specifically, Nominatim's `ISO3166-2-lvl6` returns the *province* code (e.g. `ES-B` for Barcelona province), but the bundle only carries the *autonomous-community* level (Catalonia). No amount of string normalization can match two different subdivision levels — #1487's fix only handles same-level names that differ in spelling.

Auditing every visited place against the shipped bundle (not just Spain) turned up the same class of failure in Belgium, Italy, Hungary, Mexico, and part of Great Britain and France too — the bundle's admin-1 granularity for those countries doesn't match what the geocoder returns, so the cached region code/name never matches any bundle feature and the region silently never highlights, even though the country-level stats are correct.

### Steps to reproduce

1. Log a trip with a place in Barcelona or Madrid, Spain.
2. Open Atlas, zoom into Spain (zoom level ≥ 6 to load the region layer).
3. Observe that Catalonia / Comunidad de Madrid never fills in as visited, even though the country-level stats correctly count places there.

### Expected behavior

A region should highlight as visited whenever a logged place's coordinates fall inside that region's bundled polygon, independent of what subdivision level the geocoder happened to report.

### Deployment method

Docker Compose

### Host OS

Synology DSM

### Accessing TREK from

Desktop browser

### Additional context

A PR is incoming — currently being worked on and tested against a real backup database (auditing all visited places, not just Barcelona/Madrid). The approach resolves a place's coordinates directly against the same bundled admin-1 polygons the client renders (point-in-polygon, the same technique already used for country-level resolution in #1331), instead of trusting the geocoder's address level — which is deterministic and guarantees a match, rather than trying to reconcile two data sources that don't share a subdivision level.

daften:
Closes #1547.

## Summary

The linked issue is about Barcelona/Madrid never highlighting their region, but the actual root cause — Nominatim's reverse-geocode address level not matching whatever admin-1 granularity the geoBoundaries bundle happens to ship per country — turned out to affect more than Spain, and more than just regions. This PR ended up fixing several related problems along the way, all stemming from the same root issue, so it's a bit bigger than the issue title suggests:

- **The region-matching bug itself** (closes #1547): `getRegionFromCoords()` resolves a place's lat/lng directly against the same bundled admin-1 polygons the client renders — offline, deterministic, and guaranteed to produce a bundle-matching code — instead of trusting Nominatim's address level. Verified against a real backup database: 25 of 118 visited places across Spain, Belgium, Italy, Hungary, Mexico and parts of GB/France were silently never highlighting before this fix; all 118 resolve correctly now.
- **A stale GB rescue path removed**: a leftover from the old Natural Earth bundle re-queried Nominatim at a finer zoom to swap a correct `GB-ENG`-style code for a county/borough code that the current geoBoundaries bundle (only 4 GB constituent countries) can never match. It was actively making things worse; removed.
- **Per-part bounding boxes for regions**: `buildRegionIndex` now stores one box per polygon part (matching how country resolution already worked), fixing the box prefilter for archipelago-style regions like Illes Balears/Canarias.
- **A border-simplification edge case**: a place on the Luxembourg side of the Sauer river fell inside Germany's simplified polygon. The place's own stored address now disambiguates this as a fallback (gated against that country's own admin0 *box*, not its exact polygon, so it doesn't reject the exact case it exists to fix) — but only tried *after* coordinates, because trying it first regressed **country** resolution too (see next point).
- **Country resolution brought in line with region resolution, for consistency**: `resolveCountryCode(Sync)` used to try the place's *address* first, falling back to coordinates. `getCountryFromAddress()`'s "2-letter uppercase last segment = ISO code" heuristic collides with a long list of US state abbreviations that are also real ISO country codes (DE=Germany, GA=Georgia, IN=India, LA=Laos, MD=Moldova, MA=Morocco, MN=Mongolia, MO=Macau, MT=Malta, NE=Niger, PA=Panama, SD=Sudan, TN=Tunisia, VA=Vatican, CA=Canada, ...). A place stored as "..., San Francisco, CA" was resolving to Canada — confirmed live on a real account (`getStats()` was showing "Canada, 0 places"). Country resolution now uses the exact same coordinates-first, address-as-gated-fallback order as region resolution, so the two code paths behave consistently instead of one being more correct than the other.

## Why coordinates-first, not address-first

Nominatim's address level and the bundle's admin-1 granularity frequently disagree (a province code vs. an autonomous-community-level bundle, for example) in ways no amount of string normalization can bridge. Resolving directly against the same polygons the client renders sidesteps the whole class of bug — the stored code is then guaranteed to equal a bundle feature — and it's also strictly faster: the common case becomes an in-memory box + point-in-polygon test instead of a Nominatim HTTP round-trip throttled to ≥1.1s between calls.

## Testing

- Full server suite green (3798 tests).
- Full client suite green (3189 tests).
- Verified against a real backup database (118 places across ~25 countries): 118/118 now resolve to a bundle-matching region, up from ~93.
- New/updated unit tests cover: the GB rescue removal, per-part region boxes, the Luxembourg border case, the address-fallback plausibility gate, and the US state-abbreviation regression for both region and country resolution.

A separate, smaller follow-up PR adds the ability to hide a wrongly-highlighted region (with a country-level cascade) — not a bug fix, so kept out of this one.

### DIFF
--- a/server/src/db/migrations.ts
+++ b/server/src/db/migrations.ts
@@ -3660,6 +3660,25 @@ function runMigrations(db: Database.Database): void {
         console.warn('[migrations] Non-fatal migration step failed:', err);
       }
     },
+
+    // `place_regions` is a re-derivable Nominatim cache, only ever populated for a place ID
+    // that isn't already cached — so a wrong row, once written, was permanent. Region
+    // resolution now resolves a place's lat/lng directly against the bundled admin1 polygons
+    // (the same ones the client renders) instead of trusting Nominatim's address level, which
+    // could name a subdivision level the bundle doesn't carry (Barcelona's ES-B province vs
+    // the bundle's ES-CT autonomous community) and never highlight. That fix only helps
+    // places re-resolved after it, so clear the cache once and let every place re-resolve on
+    // the next Atlas load. The country_code stored alongside is cleared too, which also drops
+    // the old wrong-country rows a US-state-abbreviation address used to produce.
+    () => {
+      try {
+        db.exec('DELETE FROM place_regions');
+      } catch (err) {
+        // place_regions is created by an earlier migration; tolerate its absence on an
+        // unusual partial DB rather than aborting startup.
+        if (!(err instanceof Error) || !err.message.includes('no such table')) throw err;
+      }
+    },
   ];
 
   if (currentVersion < migrations.length) {
--- a/server/src/services/atlasService.ts
+++ b/server/src/services/atlasService.ts
@@ -584,6 +584,38 @@ function pointInGeometry(lng: number, lat: number, geom: CompactGeom): boolean {
 type CompactGeom = { rings: Float64Array[]; polyRingCounts: number[] };
 type Box = [number, number, number, number]; // [minLng, minLat, maxLng, maxLat]
 
+// Flatten a GeoJSON Polygon/MultiPolygon into the same compact form the country index uses,
+// plus one bounding box per part (a part-box, like buildCountryIndexes builds, so an
+// archipelago-style region — Illes Balears, Canarias, … — gets one tight box per island
+// group rather than one box spanning the whole span between them). Used to resolve admin1
+// regions against the bundle, which are parsed per country on demand rather than held whole.
+function compactGeomFromGeometry(geometry: { type: string; coordinates: unknown }): { geom: CompactGeom; boxes: Box[] } {
+  const parts = (geometry.type === 'Polygon' ? [geometry.coordinates] : geometry.coordinates) as number[][][][];
+  const rings: Float64Array[] = [];
+  const polyRingCounts: number[] = [];
+  const boxes: Box[] = [];
+  for (const part of parts) {
+    polyRingCounts.push(part.length);
+    for (const ring of part) {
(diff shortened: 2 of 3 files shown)


jubnl:
Hey, can you please update your fork so it is in sync with the TREK repo ?

daften:
Hi, I rebased this on top of dev. I was a bit confused since previous PR's got a comment indicating the commits needed to be cherry picked in a release branch.

Issue #1490, opened 2026-07-07
[BUG] Cant remove visited country in Atlas


lagge78:
### Pre-flight checklist

- [x] I have searched [existing issues](https://github.com/mauriceboe/TREK/issues) and this bug has not been reported yet
- [x] I am running the latest available version of TREK
- [x] I have read the [Troubleshooting guide](https://github.com/mauriceboe/TREK/wiki/Troubleshooting) and my issue is not covered there

### TREK version

3.2.1

### Describe the bug

In my atlas the country Algeria is marked as visited which I havent. I press the country and press Remove.
The country is removed. But next time I go to the Atlas just this country is marked as visited again. I have tried several times on different computers. No change.

After removal
<img width="1108" height="862" alt="Image" src="https://github.com/user-attachments/assets/3e1c46af-71e6-48fe-a008-dae5dd11082a" />

Next time I go to Atlas, this country is added again.
<img width="1142" height="812" alt="Image" src="https://github.com/user-attachments/assets/998dc360-6c60-459a-bb14-ce69f09d3072" />

### Steps to reproduce

Go to atlas
Remove Algeria
Leave Atlas
Go back and its added again,

### Expected behavior

A removed country should stay unchecked

### Deployment method

Docker Compose

### Accessing TREK from

Desktop browser

### Relevant logs or error output

```shell

```

daften:
Check out #1486, I think this might be it.

PR #1564, opened 2026-07-14
feat(atlas): let a visited region be hidden, cascading to the country when none remain

daften:
**Stacked on #1563 — this branch is built on top of `fix/atlas-region-coords-match`, so the diff currently shows all 7 commits (5 from #1563 + 2 new ones here). Check the *Commits* tab to see just the 2 this PR actually adds; the *Files changed* diff will shrink to just those once #1563 merges and this rebases/retargets cleanly onto `main`.**

Related to #1547, but not a bug fix — a small feature that came up while working that issue.

## Summary

Countries already have a hide/tombstone mechanism (`hidden_countries`, #1490): a zero-count derived country can be dismissed and stays gone across reloads. Regions had no equivalent — `unmarkRegionVisited` only ever deleted a manually-marked `visited_regions` row, a no-op for the common case of a region derived fresh from `place_regions` on every request, so there was no way to dismiss one at all.

This adds the region-level counterpart, deliberately mirroring the country-level design so the two stay consistent:

- New `hidden_regions` table, same shape as `hidden_countries`.
- `getVisitedRegions()` filters through it, the same way `getStats()` already filters through `getHiddenCountries()`.
- `unmarkRegionVisited()` now tombstones unconditionally — not just for a manually-marked region. A region with a real place attached (e.g. one misassigned by a border-simplification gap, exactly the kind of thing #1563 fixes some instances of but can't catch all of) is the actual motivating case.
- Cascade: hiding a country's last visible region also hides the country, via the existing `unmarkCountryVisited`.
- `markRegionVisited()` clears both tombstones on re-mark, so a region (and its cascade-hidden country) can come back.
- Client: clicking any visited region now offers to remove it, not just manually-marked ones. Country detail viewing remains reachable via the country search/sidebar, which was never gated on this.

One deliberate asymmetry, covered by tests: the cascade only has a *visible* effect on a country with no real place data behind it — a country with real places is never actually hidden by the tombstone (#1490's existing "reappears once it has a real place" rule), so hiding every region of a real-place-backed country leaves the country itself visible. Same existing design, just now reachable from the region side too.

## Testing

- Full server suite green.
- Full client suite green.
- New tests: server-side hide/cascade/re-mark behavior (including the real-place-vs-manual asymmetry above), client-side region-click flow for a place-derived (non-manually-marked) region.

### DIFF
--- a/client/src/pages/AtlasPage.test.tsx
+++ b/client/src/pages/AtlasPage.test.tsx
@@ -1578,6 +1578,61 @@ describe('AtlasPage', () => {
     });
   });
 
+  describe('FE-PAGE-ATLAS-048: clicking a visited (place-derived, not manually-marked) region opens the hide-region popup', () => {
+    it('offers to remove a region that came from real place data, not just a manually-marked one', async () => {
+      // FR keeps real tripCount/placeCount from atlasStatsResponse so the country layer's
+      // own click handler (only active for a zero-count country) stays a no-op — this test
+      // is only about the region layer. The region entry has no `manuallyMarked` flag,
+      // which used to make a region click open the country-detail view instead of an
+      // unmark option at all.
+      server.use(
+        http.get('/api/addons/atlas/regions', () => HttpResponse.json({
+          regions: { FR: [{ code: 'FR-IDF', name: 'Île-de-France', placeCount: 3 }] },
+        })),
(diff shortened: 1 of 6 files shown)


jubnl:
Hey, can you please update your fork so it is in sync with the TREK repo ?

daften:
Done!

PR #1572, opened 2026-07-16
fix(budget): re-split expenses when a member is removed

### LINKED ISSUE #1553: [BUG] Removed Guest Members stay in Transactions
### Pre-flight checklist

### TREK version

3.3.0

### Describe the bug

I have a trip with 1 real member (myself) and 7 guest members. 
I just created a lot of transactions split 8 ways between all the travelers.

I then removed 4 of those 8 travelers, because i would rather split the cost between Families than single persons. But now my transactions are still split 8 ways, and i am unable to deselect the removed guest members from the split:

Image

The calculation is also incorrect now, since its still being split into 8 instead to 4.
Even if I deselect all Guest members now, it still tries to split it 4 ways (to the already deleted guest members)

### Steps to reproduce

1. Create Guest Members
2. Create Transactions between multiple Guest Members
3. Delete Guest Members
4. Transaction Split is now incorrect and throws an Error when trying to change it.

### Expected behavior

When deleting guest members, transactions should be recalculated with the split between the remaining members

### Deployment method

Docker Compose

### Host OS

Debian

### Accessing TREK from

Desktop browser

### Browser (if applicable)

Firefox

```shell
[ERROR] 2026-07-14T10:43:01 PUT /api/trips/1/budget/11 500 2ms ip=192.168.1.1

Unhandled error: SqliteError: FOREIGN KEY constraint failed

    at Object.updateBudgetItem (/app/server/dist/services/budgetService.js:234:20)

    at BudgetService.update (/app/server/dist/nest/budget/budget.service.js:81:20)

    at process.processTicksAndRejections (node:internal/process/task_queues:104:5)

    at async BudgetController.update (/app/server/dist/nest/budget/budget.controller.js:127:25) {

  code: 'SQLITE_CONSTRAINT_FOREIGNKEY'
```

### Screenshots

_No response_

### Additional context

_No response_

subdee:
## Description

Removing a member (a guest, or a full account) from a trip left their expense
splits half-cleaned. `budget_item_members` rows cascade away with the deleted
`users` row, but `budget_items.persons` — the denormalized divisor the per-person
column uses — kept the old count, so an 8-way split kept dividing by 8 after four
people were removed. And because a client that loaded before the deletion still
sent the removed ids on the next save, the update hit `SQLITE_CONSTRAINT_FOREIGNKEY`
(a 500) — `INSERT OR IGNORE` does not suppress FK violations.

This PR:
- Re-derives `persons` from the surviving members before the `users` row is
  deleted, in both `deleteGuest` and account cleanup.
- Drops user ids with no `users` row from every budget member/payer write, so a
  save from a stale client succeeds (dropping the ghost) instead of 500ing.
- Refetches budget items after a guest deletion on the client, which also
  refreshes the Settle Up card (it re-queries whenever budget items change).

Settle Up / balances math needed no change — it always derives from live
member/payer rows, never from `persons`; the stale card was purely a missing
client refetch.

Closes #1553

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Checklist

### DIFF
--- a/client/src/components/Trips/TripMembersModal.tsx
+++ b/client/src/components/Trips/TripMembersModal.tsx
@@ -290,6 +290,7 @@ export default function TripMembersModal({ isOpen, onClose, tripId, tripTitle, o
   const { t } = useTranslation()
   const can = useCanDo()
   const trip = useTripStore((s) => s.trip)
+  const loadBudgetItems = useTripStore((s) => s.loadBudgetItems)
   const canManageMembers = can('member_manage', trip)
   const canManageShare = can('share_manage', trip)
 
@@ -387,6 +388,7 @@ export default function TripMembersModal({ isOpen, onClose, tripId, tripTitle, o
     try {
       await tripsApi.deleteGuest(tripId, userId)
       await loadMembers(true)
+      await loadBudgetItems(tripId)
       toast.success(t('members.guestRemoved'))
     } catch {
       toast.error(t('members.removeError'))
--- a/server/src/services/budgetService.ts
+++ b/server/src/services/budgetService.ts
@@ -30,13 +30,23 @@ function loadItemPayers(itemId: number | string) {
   return rows.map(p => ({ ...p, avatar_url: avatarUrl(p) }));
 }
 
+function knownUserIds(userIds: number[]): Set<number> {
+  const unique = [...new Set(userIds)];
+  if (unique.length === 0) return new Set();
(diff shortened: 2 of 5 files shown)


mauriceboe (approved):
LGTM — clean fix, diagnosis is spot on and the tests fail without it. Thanks Konstantinos! 🙌

PR #1581, opened 2026-07-17
fix(packing): apply templates into the active list tab

### LINKED ISSUE #1565: [BUG] Packing list template can only be applied to "Shared" tab
### Pre-flight checklist

- I have searched [existing issues](https://github.com/mauriceboe/TREK/issues) and this bug has not been reported yet
- I am running the latest available version of TREK
- I have read the [Troubleshooting guide](https://github.com/mauriceboe/TREK/wiki/Troubleshooting) and my issue is not covered there

### TREK version

3.3.0

### Describe the bug

When using the Apply Template feature in Lists > Packing List, the items are added to the "Shared" tab even if "My List" tab is active / selected

### Steps to reproduce

1. Select a trip
2. Click "Lists" below the header
3. Under Packing List, click "My List"
4. Click on the Apply Template button and select a template
5. All the categories / items from the template are added to the "Shared" tab

### Expected behavior

I would expect the categories & items to be added to the currently active tab, either "Shared" or "My List"

### Deployment method

Docker Compose

### Host OS

N/A

### Accessing TREK from

Desktop browser

### Browser (if applicable)

Zen 1.21.6b

```shell
```

### Screenshots

_No response_

### Additional context

_No response_

subdee:
## Description

Applying a packing template always added the items to the **Shared** tab, even with **My List** active.

Two things were broken on the same path. `applyTemplate()` in `server/src/services/packingService.ts:393` inserted rows *omitting* `is_private` and `owner_id`, so SQLite applied the column default `is_private INTEGER NOT NULL DEFAULT 0` (`server/src/db/migrations.ts:3140`) — the Common tier. The function took no visibility or owner argument at all, so it could not produce a personal item. The active tab could not have reached it in any case: the client posted no body (`client/src/api/client.ts:452`) and the route dropped the acting user (`server/src/nest/packing/packing.controller.ts:355`).

The tab state also lived in the wrong place for the button users actually press. `TripPlannerPage` renders the panel with `inlineHeader={false}`, and the in-panel dropdown is gated on `inlineHeader &&`, so the live control is the standalone `ApplyTemplateButton` — rendered *outside* the panel, with no access to its `view` state. Passing a parameter down was not enough; the state had to be lifted.

The fix threads the active view through the whole path: `view` moves up to `ListsContainer` (the common parent of the panel and the button), the panel takes it as a controlled prop and falls back to its own state, and the service gains `visibility` + `ownerId`. The service default stays `'common'`, so the existing MCP caller (`server/src/mcp/tools/packing.ts:305`) is unchanged.

Two deliberate decisions worth review:

- **Common items stay unowned.** Stamping the applier as owner would let `setItemSharing` lock every other member out of re-sharing them (it treats a null owner as claimable). That regression isn't part of this bug, so Common behaviour is byte-for-byte what it was.
- **The broadcast is now privacy-aware** (`broadcastItem` rather than room-wide `broadcast`), so a personal apply is not pushed to the whole trip. The client currently has no `case` for `packing:template-applied` in `remoteEventHandler.ts`, so nothing renders from it today either way — but the payload should not carry another member's personal items onto the wire.

Layering: the behaviour lives in `services/` (the Nest service stays a pass-through), the request shape is in the `shared/` Zod contract, and the tab state sits in the page container that owns both components.

Out of scope, found while here and worth separate issues: `packing:template-applied` has no `case` in `remoteEventHandler.ts` (other members only see templated items after a reload), and Apply Template has no repo/offline path — it's online-only and writes to the store without a Dexie write-through, same as `bulkImport`.

Closes #1565

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Checklist
- [x] I have read the [Contributing Guidelines](https://github.com/mauriceboe/TREK/wiki/C


jubnl (approved):
lgtm

PR #1516, opened 2026-07-11
Feat:LDAP Auth v2

AlexBander:
## Description
Adds native LDAP authentication against FreeIPA and OpenLDAP servers. Users not found in LDAP fall back to local authentication, so existing local admin accounts keep working without changes.

- RFC-4515 username escaping to prevent LDAP injection
- Role mapping via LDAP group membership (admin / user)
- Optional access control via LDAP_ALLOWED_GROUP — admins bypass this check
- TLS support with custom CA certificate (LDAP_TLS_CA)
- Graceful fallback to local login when LDAP is unavailable or user not found
- i18n strings for all 22 supported languages
- Unit tests for bind, group membership, role assignment and injection escaping

New environment variables:
  LDAP_URL            ldaps://ipa.example.com:636
  LDAP_BIND_DN        service account DN
  LDAP_BIND_PW        service account password
  LDAP_BASE           search base DN
  LDAP_ADMIN_GROUP    members get role=admin
  LDAP_ALLOWED_GROUP  only members may log in (admins exempt)
  LDAP_TLS_CA         path to CA certificate file

## Related Issue or Discussion
[#1035](https://github.com/mauriceboe/TREK/pull/1035)

## Type of Change
- [ ] Bug fix
- [X] New feature
- [ ] Breaking change
- [ ] Documentation update

## Checklist
- [X] I have read the [Contributing Guidelines](https://github.com/mauriceboe/TREK/wiki/Contributing)
- [X] My branch is [up to date with `dev`](https://github.com/mauriceboe/TREK/wiki/Development-environment#3-keep-your-fork-up-to-date)
- [X] This PR targets the `dev` branch, not `main` *(wiki-only PRs are exempt)*
- [X] I have tested my changes locally
- [X] I have added/updated tests that prove my fix is effective or that my feature works
- [X] I have updated documentation if needed

PR #1623, opened 2026-07-20
feat(auth): LDAP/LDAPS authentication with admin UI and login page selector based v3.4.x

AlexBander:
## Description
Adds native LDAP authentication support for FreeIPA and OpenLDAP with
role mapping, access control, and a configurable login page experience as defined in the .env file.

Backend:
- ldapService.ts: bind/search/group-check, RFC-4515 injection escaping
- ldapLoginUser() async wrapper in authService — falls back to local
  login when user not found in LDAP or LDAP is unconfigured
- isOidcOnlyMode() respects LDAP_URL so LDAP users can log in even
  when password login is disabled
- ldap_configured and ldap_default_method exposed via getAppConfig()
- ldap_default_method persisted in app_settings

Admin UI:
- New setting ldap_default_method (ldap / local / both) controls which
  method is pre-selected on the login page

Login page:
- Tab toggle between LDAP and local when both methods are available
- Email field switches to username field for LDAP login
- LDAP uid extracted from email input automatically

Tests:
- 9 unit tests for bind, group membership, role assignment and
  LDAP injection escaping (all passing)

i18n:
- login.ldap.* and admin.ldap.* keys for all supported languages (via AI-gen)

New environment variables:
  LDAP_URL             ldaps://ipa.example.com:636
  LDAP_BIND_DN         service account DN
  LDAP_BIND_PW         service account password
  LDAP_BASE            search base DN
  LDAP_ADMIN_GROUP     members get role=admin
  LDAP_ALLOWED_GROUP   only members may log in (admins exempt)
  LDAP_TLS_CA          path to CA certificate file

[#1034](https://github.com/mauriceboe/TREK/pull/1034)
[#1035](https://github.com/mauriceboe/TREK/pull/1035)
[#1515](https://github.com/mauriceboe/TREK/pull/1515)
[#1516](https://github.com/mauriceboe/TREK/pull/1516) (withdrew)
[#1530](https://github.com/liketrek/TREK/pull/1530) (withdrew)

## Type of Change
- [ ] Bug fix
- [X] New feature
- [ ] Breaking change
- [ ] Documentation update

## Checklist
- [X] I have read the [Contributing Guidelines](https://github.com/mauriceboe/TREK/wiki/Contributing)
- [X] My branch is [up to date with `dev`](https://github.com/mauriceboe/TREK/wiki/Development-environment#3-keep-your-fork-up-to-date)
- [X] This PR targets the `dev` branch, not `main` *(wiki-only PRs are exempt)*
- [X] I have tested my changes locally
- [X] I have added/updated tests that prove my fix is effective or that my feature works
- [X] I have updated documentation if needed

### DIFF
--- a/.github/workflows/lint-prettier.yml
+++ b/.github/workflows/lint-prettier.yml
@@ -1,7 +1,7 @@
 name: Lint & Prettier
 on:
   pull_request:
-    branches: [main, dev]
+    branches: [main, dev, dev-test]
 jobs:
   lint:
     runs-on: ubuntu-latest
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -5,7 +5,7 @@ permissions:
 
 on:
   pull_request:
-    branches: [main, dev]
+    branches: [main, dev, dev-test]
     paths:
       - 'server/**'
       - 'client/**'
--- a/.gitignore
+++ b/.gitignore
@@ -31,6 +31,8 @@ server/uploads/
 # OS files
 .DS_Store
 Thumbs.db
+*.orig
+*.rej
 
 # IDE
 .vscode/
--- a/.npmrc
+++ b/.npmrc
@@ -1,0 +1,1 @@
+registry=https://registry.npmjs.org/
--- a/client/src/pages/LoginPage.tsx
+++ b/client/src/pages/LoginPage.tsx
@@ -65,6 +65,7 @@ export default function LoginPage(): React.ReactElement {
     noRedirect,
     showRegisterOption,
     oidcOnly,
+    authMethod, setAuthMethod,
     handleDemoLogin,
     handleSubmit,
     handlePasskeyLogin,
@@ -774,7 +775,29 @@ export default function LoginPage(): React.ReactElement {
                         : t('login.subtitle')}
                 </p>
 
-                <form onSubmit={handleSubmit} style={{ display: 'flex', flexDirection: 'column', gap: 16 }}>
+                {/* LDAP / Local toggle */}
+            {appConfig?.ldap_configured && appConfig?.ldap_default_method === 'both' && mode === 'login' && !mfaStep && !passwordChangeStep && (
+              <div style={{ display: 'flex', borderRadius: 10, border: '1px solid #e5e7eb', overflow: 'hidden', marginBottom: 4 }}>
+                {(['ldap', 'local'] as const).map((m) => (
+                  <button
+                    key={m}
+                    type="button"
+                    onClick={() => setAuthMethod(m)}
+                    style={{
+                      flex: 1, padding: '8px 0', border: 'none', cursor: 'pointer',
+                      fontFamily: 'inherit', fontSize: 'calc(13px * var(--fs-scale-body, 1))', fontWeight: 600,
+                      background: authMethod === m ? '#111827' : 'white',
+                      color: authMethod === m ? 'white' : '#6b7280',
+                      transition: 'background 0.15s, color 0.15s',
+                    }}
+                  >
+                    {m === 'ldap' ? t('login.ldap.method') : t('login.local.method')}
+                  </button>
+                ))}
+              </div>
+            )}
+
+            <form onSubmit={handleSubmit} style={{ display: 'flex', flexDirection: 'column', gap: 16 }}>
                   {error && (
                     <div
                       style={{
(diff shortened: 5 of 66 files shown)


mauriceboe (changes requested):
Thanks for putting this together — LDAP has been asked for more than once, and the RFC 4515 escaping in `ldapService.ts` is done properly, with a test that actually proves it. Keeping the bind credentials in env vars, out of the DB and out of `getAppConfig()`, is right too.

I can't take it as it stands. The most serious problems are outside the LDAP path itself:

**`ldapts` is in the wrong manifest — this breaks every login, LDAP or not.** It's declared in the root `package.json`, but the production image installs the server workspace only (`npm ci --workspace=server --omit=dev`, Dockerfile:52), so the module never lands in the image. Since login now goes through `ldapLoginUser()`, which does `await import('./ldapService')` unconditionally (`authService.ts:545`), that import throws on *every* login — on instances with no LDAP configured at all. It needs to move to `server/package.json`, and the import should sit behind the `getLdapConfig()` check.

**MFA is skipped.** `loginUser()` returns `{ mfa_required, mfa_token }` when `user.mfa_enabled` is set (`authService.ts:512`). `ldapLoginUser()` never reads that flag and calls `generateToken(user)` straight away, so any account reachable through LDAP signs in with one factor — including where `require_mfa` is enforced.

**Existing roles get overwritten.** Every LDAP login writes the directory role onto the matched account. With `LDAP_ADMIN_GROUP` unset, `isAdmin` is always false, so the first LDAP login by an existing admin demotes them to `user`. On a single-admin instance that locks the instance out of its own settings.

**Directory users without `mail` share one account.** `ldapService.ts:105` falls back to `String(entry['mail'] || '')` and the lookup is `WHERE LOWER(email) = LOWER('')`, so the second such user matches the first one's row and gets a session on their account.

**OIDC-only mode is switched off by setting `LDAP_URL`.** The guard became `isOidcOnlyMode() && !process.env.LDAP_URL` (`authService.ts:474`), and `ldapLoginUser` falls back to `loginUser(body)` whenever the directory doesn't know the user — so local bcrypt passwords work again on instances where the admin deliberately turned password login off.

**The admin setting is never saved.** `AdminSettingsTab.tsx:208` PATCHes `/api/admin/settings`, which doesn't exist — every other setting on that tab goes through `adminApi.*`. The 404 is swallowed, so the radio looks like it works.

**The login form can't accept an LDAP username.** The input is still `type="email" required` while the client truncates at `@`. A uid that isn't a mail local-part can't be submitted, and with `ldap_default_method: 'ldap'` the toggle isn't rendered at all, so local accounts — bootstrap admin included — have no way in.

Smaller ones: no `connectTimeout`/`timeout` on the client, and the directory round trip runs before any local check, so an unreachable server hangs every login; `remember_me` is dropped (the signature reads `remember`, and `generateToken` is called without the flag); the lookup misses the `COALESCE(is_guest, 0) = 0` filter every other auth path applies (#1362); and `ldap_default_method` is client-side only — the server always tries LDAP first and always falls back to local, so "LDAP only" and "Local only" don't do what the labels say.

Could you also drop what isn't part of the feature? The `dev-test` CI triggers in `test.yml`/`lint-prettier.yml`, the new `.npmrc`, the `*.orig`/`*.rej` gitignore entries, and the `cp -r` appended to the root `build` script (POSIX-only, and the Dockerfile already does that copy).

The branch is also ~90 commits behind `dev` by now, so it'll want a rebase either way.

Happy to review again once these are sorted. If I haven't heard back by **29 July** I'll close this for the time being — that's not a rejection of the feature, and reopening is easy whenever you have time to pick it back up.

mauriceboe:
One more thing, and this one is on us rather than on you: it wasn't written down anywhere you could have read it.

The backend is mid-migration to NestJS. `server/src/nest/` is where new server code belongs — a service, a controller and a module per domain, registered in `app.module.ts`, with the prefix added to `DEFAULT_NEST_PREFIXES` in `strangler.ts`. `server/src/services/` and `server/src/routes/` are the legacy Express layer that migration is supposed to shrink. This PR adds `server/src/services/ldapService.ts` plus the login wiring in `authService.ts`, which grows exactly the layer we're trying to retire.

The spec is `server/src/nest/README.md` — module layout, the recipe for moving a `src/services/*` service across, and the three kinds of tests each module ships (unit, parity, e2e). `weather/` is the reference implementation worth copying. One detail that matters for LDAP specifically: inside a Nest module you don't read `process.env` directly (ESLint enforces it there) — config comes from `app-config` via a `registerAs` token, or `RuntimeEnvService`/`readEnv()` for values an operator can flip at runtime. That's a better home for `LDAP_URL` and friends than reading env vars in the service.

I know that's a bigger ask on top of the correctness fixes. If you want to take it on, say so and we'll sort out the timing — the date I mentioned isn't meant to push you off something you're actively working on. If you'd rather hand it over, that's completely fine too; the work you've done on the escaping and the group logic is worth keeping either way.

PR #1625, opened 2026-07-20
feat(backup): mirror backups to an external S3-compatible target

fbnlrz:
## Description

TREK's backups have only ever lived on the same volume as the data they protect,
so a single lost disk takes both. This adds an optional S3-compatible target that
every backup — manual *and* scheduled — is mirrored to.

The local archive is always kept. The remote copy is pushed *in addition*, so an
unreachable target costs you the off-box copy, never the backup itself.

**Scope is deliberately backups-only.** The uploads/object-storage half of the
discussion (#275) is not touched here.

### What's in it

- **One post-write hook both builders call.** `createBackup()` and the
  scheduler's `runBackup()` have always diverged; routing both through
  `onBackupWritten()` means manual and automatic backups get identical off-box
  coverage. The call sits *outside* the `try` that deletes the archive on
  failure — a failing target must never take the local copy with it.
- **A `BackupTarget` interface**, which is the seam @jubnl suggested in #228.
  S3 is the only implementation today; a further backend plugs in without
  touching either builder. The unprefixed `S3_*` / `STORAGE_BACKEND` namespace is
  deliberately left free for the general storage backend discussed there.
- **Config in `app_settings`, SMTP-style.** The secret access key is encrypted
  with `apiKeyCrypto` and never returned to the browser — a read reports whether
  one is set and echoes a mask that a save round-trips. The new encrypted key is
  registered in `migrate-encryption.ts` in the same commit, or key rotation would
  silently orphan it.
- **`BACKUP_S3_*` env vars** take precedence and lock the admin form, matching how
  `SMTP_PASS` overrides the stored `smtp_pass`.
- **Test connection checks writes, not just reachability.** `HeadBucket`, then a
  probe object written and deleted. A read-only key therefore fails the test
  instead of passing it and then breaking every backup silently.
- **The backup list merges disk and target**, badging where each archive lives.
  An S3-only archive can be restored (fetched, then run through the existing
  integrity / zip-slip / zip-bomb guards) and deleted. An unreachable bucket
  degrades to the local list rather than failing the call.
- **Deleting removes both copies**, and reports a target that refused.
- **"Upload all existing backups"** backfills what predates the target, skipping
  archives already present.

### Security notes

- The endpoint is validated with `checkSsrf` before any request, respecting
  `ALLOW_INTERNAL_NETWORK` because a LAN MinIO/Garage is the normal self-hosted
  case. The AWS SDK does its own networking and does not route through TREK's
  DNS-pinned `safeFetch`, so this is a config-time check. Pinning is reachable
  via a custom `requestHandler` but is its own change; the module says so rather
  than overclaiming.
- Endpoints that serve the S3 API under a path (Supabase Storage, or a Ceph RGW /
  MinIO behind an ingress) are passed through unchanged.
- The secret never reaches logs or the audit log — only whether one was supplied.

### Also fixed here

`wiki/Backups.md` claimed "Not included: the encryption key". That stopped being
true in 3.1.0 (ad893eb1), which bundles `.encryption_key` whenever
`ENCRYPTION_KEY` is not set as an env var. It matters more once archives can be
pushed to third-party storage, so the page and the panel now say what a backup
actually contains.
 **All six deployment paths** carry the new variables: `.env.example`,
  `docker-compose.yml`, the Helm `values.yaml` + `secret.yaml`, the Unraid
  template and `wiki/Environment-Variables.md`.
### Deliberately not in this PR

Optional AES-256 encryption of the archive before transmission (#228), remote
retention pruning, and multiple named destinations (#373). Each is its own
change.

Addresses discussion #228, and the backup half of #373 and #1112.

## Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [x] Documentation update

## Checklist


client/src/components/Admin/BackupPanel.tsx
subdee: You are returning the result of the remote backup execution so you can handle any errors but this is not being caught here. You only catch local backup errors. You should add a check for remote success here before you display the success toast, otherwise the user is being misinformed.

subdee (changes requested):

mauriceboe:
hey @fbnlrz,

thanks for this, genuinely solid work (the s3mini choice over the aws sdk, the streaming, the "a failing remote never kills the local backup" invariant + tests). this is NOT a "bad code" reject.

honestly it set off a pretty heated internal debate 😅.. but where we landed is it's a scope thing, not a code thing. we want to do storage properly across the whole app: one general layer everything goes through (uploads, images, docs, backups) with pluggable backends + named instances and routing. a backup-only BACKUP_S3_* setup would lock us into a contract we'd carry forever or break later.

Julien drafting specs for that now, and your BackupTarget interface is a great starting point.

gonna close this as "superseded by the storage work" rather than merge.. really appreciate the effort, sorry for the back and forth.🙏

PR #1632, opened 2026-07-20
Feat/journey external picture provider rework

SimMesg20:
## Description

This feature makes it possible to directly select pictures from external providers, when adding a picture to a day or place.
This is to reduce the extra step of first having to add all external pictures to the gallery.

image

https://github.com/liketrek/TREK/discussions/1614

## Type of Change
- [x] New feature
- [x] Documentation update

## Checklist


mauriceboe (changes requested):
Nice work.

Two things before I merge: the proximity sort never actually gets coordinates — `searchPhotos` doesn't send `withExif: true` (the album path does), and the Synology search doesn't ask for `gps` in `additional`. The Immich one predates your PR, but together it means "nearby first" is a no-op right now. Second: if the provider-photo call fails after a new entry was created, the editor stays open with `entry.id === 0`, so saving again creates a duplicate entry.

SimMesg20:
> Nice work.
> 
> Two things before I merge: the proximity sort never actually gets coordinates — `searchPhotos` doesn't send `withExif: true` (the album path does), and the Synology search doesn't ask for `gps` in `additional`. The Immich one predates your PR, but together it means "nearby first" is a no-op right now. Second: if the provider-photo call fails after a new entry was created, the editor stays open with `entry.id === 0`, so saving again creates a duplicate entry.

Ill look at it after work and fix it up, thanks for the feedback

mauriceboe (approved):

PR #1649, opened 2026-07-21
fix(settings): let an empty user value fall back to the admin default

### LINKED ISSUE #1634: [BUG] system-wide mapbox token doesn't work
### Pre-flight checklist

### TREK version

3.4.1

### Describe the bug

I am not sure if I am missing a step, however, I can't have a system-wide mapbox token.

When setting the mapbox token as admin, users still need to provide their own token. The admin mapbox token is either not set as system-wide or is ignored.

Maybe, it doesn't work because the user already set a token at some point? Even deleting it, the user can't fallback to the admin token.

### Steps to reproduce

1. Set the mapbox token in the admin console as user defaults
2. Go to the map settings under the user settings
3. Choose Mapbox GL as map provider
4. Leave the mapbox access token input field empty
5. Mapbox won't load for a user

### Expected behavior

Defining the mapbox token in the user defaults in the admin console, makes it available to all users. The user just needs to select mapbox as map provider and leave the token input empty. Then the admin token is automatically used and the user has access to a mapbox map.

### Deployment method

Docker Compose

### Host OS

_No response_

### Accessing TREK from

Desktop browser

### Browser (if applicable)

_No response_

```shell

```

### Screenshots

_No response_

### Additional context

_No response_

fbnlrz:
### Problem

Setting a system-wide Mapbox token in the admin console's *user defaults* had no effect for most users. The admin-defaults merge (`{ ...adminDefaults, ...userSettings }`) kept any user value that was merely *present*, and the client's Settings save always writes every field — a blank Mapbox token included. So any user who ever opened Map Settings had an empty-string `mapbox_access_token` row that shadowed the admin default, and clearing their own token could never restore the fallback.

A brand-new user (no settings rows) inherited the default fine; only existing users were affected — which matches the report.

### Fix

In `getUserSettings`, treat an empty/null user value for a **defaultable** key as "not set" when an admin default exists, so it falls through to the admin/system-wide default. Non-defaultable keys, and defaultable keys without an admin default, keep their exact stored value (no regression).

Single server-side fix point — covers desktop + mobile and every defaultable key, not just the Mapbox token.

### Tests

Five new `settingsService` cases: new-user inheritance, empty-value fall-through, non-empty override, empty-with-no-default stays empty, and a non-defaultable empty value is preserved.

Fixes #1634

### DIFF
--- a/server/src/services/settingsService.ts
+++ b/server/src/services/settingsService.ts
@@ -36,6 +36,8 @@ export const DEFAULTABLE_USER_SETTING_KEYS = [
 
 type DefaultableKey = typeof DEFAULTABLE_USER_SETTING_KEYS[number];
 
+const DEFAULTABLE_USER_SETTING_KEY_SET = new Set<string>(DEFAULTABLE_USER_SETTING_KEYS);
+
 const VALID_VALUES: Partial<Record<DefaultableKey, unknown[]>> = {
   temperature_unit: ['fahrenheit', 'celsius'],
   distance_unit: ['metric', 'imperial'],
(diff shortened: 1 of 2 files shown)


subdee (dismissed):
There is a pre-existing `llm_api_key` leakage that used to be only for new users but now could potentially be exposed to existing users that don't set the key and inherit the global one. Might be good to address in a separate issue.

mauriceboe (approved):
Reviewed, and added masking for inherited encrypted admin defaults — a shared llm_api_key falling through to a user must never reach the client in cleartext (addresses the review note). The empty-fallback logic is right; server tests green including the two new masking cases.

Issue #1535, opened 2026-07-12
[BUG] Layover countries marked visited when flights are imported from AirTrail


steffenrapp:
### Pre-flight checklist

- [x] I have searched [existing issues](https://github.com/mauriceboe/TREK/issues) and this bug has not been reported yet
- [x] I am running the latest available version of TREK
- [x] I have read the [Troubleshooting guide](https://github.com/mauriceboe/TREK/wiki/Troubleshooting) and my issue is not covered there

### TREK version

3.3.0

### Describe the bug

When importing flights from AirTrail they cannot be connected to multi-leg flights (see #1431). So stops cannot be declared as layover. As a consequence these countries get saved as visited which is not correct (see #1366 and #1486).

### Steps to reproduce

1. Import flights from AirTrail with layover
2. Country with layover gets saved as visited in Atlas and can’t be removed.

### Expected behavior

Flights from AirTrail should be able to get joined so layover countries don’t count as visited.

### Deployment method

Docker Compose

### Accessing TREK from

Desktop browser

### Relevant logs or error output

```shell
```

steffenrapp:
@mauriceboe I just noticed that it seems that this is not fixed. Layover countries are still marked as visited when flights are imported as multi-leg from AirTrail. Can you reopen the issue or should I create a new one?

mauriceboe:
reopened

mauriceboe:
Fixed for 4.0.0.

The earlier fix only covered a layover stored as a stop inside one booking. Joining AirTrail legs into one booking is opt-in, so legs imported separately store the connection airport as the arrival of one booking and the departure of the next, which that filter cannot tell apart from two real arrivals. That is why it looked unfixed.

Atlas now treats an arrival and a departure at the same airport within 24 h of each other as a connection, so it applies to the flights already in your database without re-importing anything. A return flight home is excluded, including when it lands at a different airport of the same city.

steffenrapp:
Sorry I think this needs to be reopened again. I just tested it on v4.2.1 and still layover countries are displayed as visited in Atlas. In one example there's a layover time of 4h. The flights have been imported from AirTrail and joined as one booking.

mauriceboe:
@subdee Can you check this out? It looks like I've failed twice already, and I don't want to mess it up a third time. :( Thanks!

subdee:
[...]

PR #1650, opened 2026-07-21
fix(airtrail): stop labelling a merged multi-leg import "removed in AirTrail"

### LINKED ISSUE #1646: [BUG] Transportation Booking shows "Not Synced" for Multi Leg flight imported from AirTrail
### Pre-flight checklist

### TREK version

3.4.1

### Describe the bug

Immediately after importing a multi-leg flight booking from AirTrail to Trek, Trek shows "Not Synced" even when no changes have been made in AirTrail

### Steps to reproduce

1. Have AirTrail integration set up properly
2. Enter a trip
3. Go to Transportation
4. Select AirTrail
5. Select a multi-leg flight and select  "import as one flight with layover"
6. Click import
7. "Not Synced" immediately shows on the saved transportation booking

### Expected behavior

Multi Leg trips should sync properly with Airtrail and update both legs as required. Should not disconnect from AirTrail just because AirTrail serves single flights and Trek allows them to be combined into multi-leg.

### Deployment method

Docker Compose

### Host OS

Unraid 7.3.1 (via Docker Compose)

### Accessing TREK from

Desktop browser

### Browser (if applicable)

Chrome

```shell

```

### Screenshots

_No response_

### Additional context

_No response_

fbnlrz:
### Problem

A multi-leg flight imported "as one flight with a layover" immediately shows **Not synced**, with the tooltip *"This flight was removed in AirTrail and no longer syncs."* — even though nothing was removed in AirTrail.

The multi-leg record is detached from sync **by design**: AirTrail has no single flight to round-trip a layover chain to, so the import creates it with `sync_enabled = 0` (see #1535, `airtrailSync.hasLocalMultiLegShape`). The badge, however, collapsed *every* `sync_enabled = 0` AirTrail row into "Not synced (removed)", so a merged booking that was never removed got a factually wrong label.

### Fix

Give the multi-leg case its own badge state. When an AirTrail reservation is unsynced **and** `metadata.legs.length > 1`, show the normal AirTrail badge with a truthful tooltip explaining it stays a one-time import. A genuinely removed single-leg flight still shows "Not synced" with the existing message.

Client-only change (the server behaviour is intentional). Adds `reservations.airtrail.layoverHint` to `en` and all locales.

### Tests

Three new `ReservationsPanel` cases: synced badge, multi-leg layover hint (no "removed" message), and single-leg removed still shows "Not synced".

Fixes #1646

### DIFF
--- a/client/src/components/Planner/ReservationsPanel.test.tsx
+++ b/client/src/components/Planner/ReservationsPanel.test.tsx
@@ -464,4 +464,33 @@ describe('ReservationsPanel', () => {
     const text = document.body.textContent || '';
     expect(text.indexOf('Mid flight')).toBeLessThan(text.indexOf('Hotel stay'));
   });
+
+  // AirTrail sync badge — three states (#1646)
+  it('FE-PLANNER-RESP-046: a synced AirTrail flight shows the AirTrail badge', () => {
+    const res = buildReservation({ title: 'Synced flight', type: 'flight', external_source: 'airtrail', sync_enabled: 1 } as any);
+    render(<ReservationsPanel {...defaultProps} reservations={[res]} />);
+    expect(screen.getByTitle('Synced from AirTrail — edits stay in sync both ways.')).toBeInTheDocument();
+    expect(screen.queryByText('Not synced')).not.toBeInTheDocument();
+  });
+
+  it('FE-PLANNER-RESP-047: a multi-leg import shows the layover hint, not the "removed" message', () => {
+    const res = buildReservation({
+      title: 'Layover flight', type: 'flight', external_source: 'airtrail', sync_enabled: 0,
+      metadata: JSON.stringify({ legs: [{ from: 'AMS' }, { from: 'IST' }] }),
+    } as any);
+    render(<ReservationsPanel {...defaultProps} reservations={[res]} />);
+    // Not falsely labelled "Not synced" / "removed in AirTrail"…
+    expect(screen.queryByText('Not synced')).not.toBeInTheDocument();
+    // …and carries the truthful layover explanation.
+    expect(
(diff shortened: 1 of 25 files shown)


nmorgoth:
### Completeness gap: the `reservation_endpoints > 2` multi-leg path is still mislabelled

The new client check keys off `metadata.legs.length > 1`, which correctly covers the #1646 case (a merged layover import always has `metadata.legs`, per `airtrailMapper.ts:258-259`).

But the server's `hasLocalMultiLegShape` (`server/src/services/airtrail/airtrailSync.ts:49-60`) detaches on a **second** criterion too:

```ts
if (Array.isArray(meta?.legs) && meta.legs.length > 1) return true;
// ...falls through to:
const row = ... 'SELECT COUNT(*) AS n FROM reservation_endpoints WHERE reservation_id = ?' ...
return row.n > 2;
```

This is the case at `airtrailSync.ts:104-107`: a user *grows* an imported single-leg flight into a multi-leg booking locally. That row ends up `sync_enabled = 0` but has **no `legs` array** in metadata — so the client's new check returns `false` and it still shows the false "Not synced / removed in AirTrail" badge. Same wrong-label class as the bug being fixed, just a narrower path.

Not blocking — the reported scenario is fully covered. But it'd be good to consciously decide on it: mirror the endpoint-count fallback on the client (if endpoint data is available on the reservation), add a code comment noting the known gap, or file a follow-up so it isn't silently lost.

mauriceboe (approved):
Reviewed, and extended the check to the endpoint-count multi-leg shape too (mirrors the server hasLocalMultiLegShape) so a locally-grown chain is not mislabelled "removed" either — closes the gap raised in review, with a test for it. Client tests green.

PR #1651, opened 2026-07-21
feat(dayplan): choose the travel mode per segment (#1281)

mauriceboe:
Raised in #1281 (and seconded in the thread by luismqoliveira / EdmisP) — the day-plan route only let you pick **one** transport mode for the whole day.

## What changes
- **Per-segment mode.** Tap the travel-time connector between two places to set that leg's mode (walking / driving / any route-provider plugin). Desktop opens a small popup at the cursor, mobile a tap menu. Each leg is routed with its own mode, so a day can mix walking, driving and plugin-routed legs — the connector shows the chosen mode's icon + time and the map redraws in it.
- **Sticky.** A per-segment choice is persisted (`day_assignments.leg_transport_mode`) and is **never overwritten** by the whole-day Foot/Car picker — that picker now sets a persisted **day default** (`days.default_transport_mode`), which a leg inherits only while it has no mode of its own. "Use day default" in the menu clears a leg's override.
- **Mobile parity.** The mobile timeline connector shows the leg's mode and is tappable, wired to the same endpoints.

## Data + API
- Two nullable columns (migration; existing itineraries keep today's single-mode behaviour), reflected in the shared DTOs.
- `PUT /trips/:id/assignments/:id/transport` and `PUT /trips/:id/days/:id/transport` — their own endpoints, so setting the day default can't wipe a day's notes/title — mirroring the `assignment_time` path.

New strings across all 23 locales.

Out of this v1 (easy follow-ups now that each leg carries its mode): per-mode map-line styling (the line stays one colour; the connector icon differentiates) and the day-total time breakdown per mode (EdmisP's request).

### DIFF
--- a/client/src/api/client.ts
+++ b/client/src/api/client.ts
@@ -27,7 +27,7 @@ import {
   type BudgetCreateItemRequest, type BudgetUpdateItemRequest,
   type PackingCreateItemRequest, type PackingUpdateItemRequest, type PackingSetSharingRequest,
   type TodoCreateItemRequest, type TodoUpdateItemRequest,
-  type AssignmentCreateRequest, type AssignmentParticipantsRequest, type AssignmentTimeRequest,
+  type AssignmentCreateRequest, type AssignmentParticipantsRequest, type AssignmentTimeRequest, type AssignmentTransportRequest,
   type PlaceBulkDeleteRequest,
   type PlaceBulkUpdateRequest,
   type DayNoteCreateRequest, type DayNoteUpdateRequest,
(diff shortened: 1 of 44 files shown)


geoida:
Testing the per-segment travel mode from this PR on a self-hosted build (dev @ c5360be2) - really nice feature. Two things after living with it: one boundary gap, and one enhancement idea.

Gap: some legs can't be given a mode

The leg-mode menu only attaches to a leg whose origin is a place-assignment, so three kinds of legs that are still drawn on the day stay read-only:

1. The first ground leg after a transport booking (ferry/train/bus arrival → next stop). The connector rendered under the booking has no menu.
2. Same for a transit booking → the following stop (bus arrival → next stop).
3. The evening accommodation bookend (last stop → the day's hotel).

Root cause (client/src/components/Planner/DayPlanSidebar.tsx):
- The clickable menu (openLegModeMenu wrapping RouteConnector) is rendered only in the place branch, gated on routeLegs[day][assignment.id].
- routeLegs is keyed by the leg's ORIGIN id (dayLegs[run[i].id]). When the origin is a transport endpoint it gets keyed by the reservation id and rendered in the transport branch as a plain RouteConnector (no menu).
- Hotel bookend legs live in the separate hotelLegs state, are always routed with the day default, and render via HotelRouteConnector (no menu).

Since leg_transport_mode is a day_assignments column ("leg leaving this stop"), legs whose origin isn't a place (booking arrival, morning hotel) have nowhere to store an override.

Suggested direction:
- Case 3 is clean and needs no schema change: the evening bookend leaves a real place, so route it with that stop's leg_transport_mode (the map already keys mode off run[i], so it's honoured there) and expose the menu on the bottom HotelRouteConnector, writing to that stop's assignment.
- Cases 1-2 need a home for the override outside day_assignments - e.g. a leg mode on the reservation, or an "incoming leg" override on the destination stop.

Happy to open a PR once you pick the direction.

Enhancement: a "public transport" option that pre-fills the transit wizard

Right now the per-segment menu covers walking / driving / cycling / plugin. It would be great to add "public transport" as a per-segment choice that, instead of drawing an OSRM line, opens the existing transit-journey search (the same flow behind create_transit_journey) pre-populated from the leg:
- origin = this stop,
- destination = the next stop,
- departure time = this stop's assignment time.

Today, to put transit on a leg you leave the segment entirely, open the transit search, and re-type from/to/time by hand. Reaching it straight from the segment menu with those three fields already filled would make mixed walking + transit days (city trips) a lot smoother, and it dovetails with the per-segment model this PR introduces.

Self-hosted note: built from dev @ c5360be2, per-segment mode otherwise works well across a 15-day multi-city trip.

PR #1672, opened 2026-07-24
fix(maps): localize map POI pin titles to the user's language

### LINKED ISSUE #1655: [BUG] Map POI pins display native names instead of English (name:en) fallback
### Pre-flight checklist

### TREK version

3.4.1

### Describe the bug

When using the map view category search (e.g. searching for "Sights", "Attractions", or "Restaurants") and clicking on pins, location titles are displayed exclusively in the local native language (e.g., Obelisco della Minerva in Rome) rather than English (Elephant and Obelisk).

While OpenStreetMap defaults to local names in the name tag, OSM POIs frequently contain localized tags such as name:en or int_name which should be prioritised for users with English language settings. This is already done when adding a place to Trek using the autocomplete box. I enjoy adding places to the list using the map which is the method I use in Wanderlog currently.

### Steps to reproduce

1. Open a trip in a non-English speaking destination (e.g., Rome, Italy).
2. Open the Map view and search for "Sights" or select a POI category overlay.
3. Click on a discovered map pin (e.g., the pin for Elephant and Obelisk at Piazza della Minerva).
4. Observe the displayed title in the pop-up/card.

### Expected behavior

The UI should attempt to display the English localized name (name:en or int_name) if present in the OSM tags, falling back to name if an English tag is not available.

### Deployment method

Unraid template

### Host OS

Unraid

### Accessing TREK from

Desktop browser

### Browser (if applicable)

Firefox

```shell

```

### Screenshots

_No response_

### Additional context

_No response_

subdee:
## Description

Map POI category search ("Sights", "Restaurants", etc.) displayed pin titles in
the native OSM `name` tag only — e.g. "Obelisco della Minerva" in Rome instead of
"Elephant and Obelisk" — even for users with a non-local language.

Root cause is two defects on the Overpass/OSM POI path, which is entirely separate
from the autocomplete path the reporter noted works correctly:

- `server/src/services/mapsService.ts:485` hard-preferred the native tag
  (`tags.name || tags['name:en'] || tags.brand`), so the native name always won and
  `int_name` was never consulted.
- The whole path was language-blind: no `lang` reached `searchOverpassPois`, unlike
  `search`/`autocomplete`, which forward the user's locale to the geocoder and get a
  localized name back.

Fix: thread the user's locale through the POI path exactly as `search`/`autocomplete`
already do — `usePoiExplore` (reads `locale`) → `mapsApi.pois` (`?lang=`) →
`@Get('pois')` → `MapsService.pois` → `searchOverpassPois(category, bbox, lang)` — and
select `name:<lang> || int_name || name` from the OSM tags (`name` as fallback, not
first choice). The language is also folded into the `POI_CACHE` key so a cached
viewport can't serve one language's titles to another. Reusing the existing
`toApiLang` normalizer maps TREK's internal codes to the right OSM subtag
(`el-GR`→`el`, `pt-BR`→`pt`). No call sites change, so desktop and mobile both get it.

Closes #1655

## Type of Change
- [x] Bug fix

## Checklist

### DIFF
--- a/client/src/api/client.ts
+++ b/client/src/api/client.ts
@@ -885,8 +885,8 @@ export const mapsApi = {
   // OSM-only POI explore: places of a category within the current map viewport bbox.
   // Overpass can be slow on a fresh (uncached) area, so this call gets a longer
   // timeout than the global default instead of aborting at 8s and showing nothing.
-  pois: (category: string, bbox: { south: number; west: number; north: number; east: number }, signal?: AbortSignal) =>
-    apiClient.get('/maps/pois', { params: { category, ...bbox }, signal, timeout: 20000 }).then(r => r.data as { pois: import('../components/Map/poiCategories').Poi[]; source: string; truncated: boolean; clamped?: boolean }),
+  pois: (category: string, bbox: { south: number; west: number; north: number; east: number }, lang?: string, signal?: AbortSignal) =>
+    apiClient.get('/maps/pois', { params: { category, ...bbox, lang }, signal, timeout: 20000 }).then(r => r.data as { pois: import('../components/Map/poiCategories').Poi[]; source: string; truncated: boolean; clamped?: boolean }),
 }
 
 export const airportsApi = {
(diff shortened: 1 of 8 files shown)


mauriceboe (approved):
Makes sense — routing the locale through the POI path the way search and autocomplete already do, and keying the cache by language so areas are not cross-served. Fallback order and tests look good, thanks!

PR #1718, opened 2026-07-26
feat(journey): splitting pdf export file into multiple files

SimMesg20:
## Description

It splits the single file export solution the Journey currently uses into multiple files.
To increase better maintainability and to offer the ability to have multiple people working on the export at the same time.
This is mostly in preperation for future changes on the export function

https://github.com/liketrek/TREK/discussions/1614
https://github.com/liketrek/TREK/discussions/1716

## Type of Change
- [x] New feature (refactor)

## Checklist
- [x] I have added/updated tests that prove my fix is effective or that my feature works (no real change in logic existing tests still pass)


jubnl (changes requested):
Discussed on discord

SimMesg20:
> Discussed on discord

changes were made, with your fix for the tests everything goes green now too

mauriceboe (changes requested):
The existing `JourneyBookPDF.test.tsx` is untouched and still passing, which is the evidence I wanted that this is a move rather than a rewrite. Thanks for keeping it that way.

Before I merge it, a few things:

**1. `const document` shadows the global.** `JourneyBookPDF.tsx:11` — inside a file that also deals with the real DOM `document`, that's a trap waiting for the next person. Please rename it (`pdfDoc` or similar).

**2. Untranslated labels.** `showJourneyExportPreview.ts:6-8` hardcodes `saveLabel`, `closeLabel` and `pagesLabel`. Either route them through `t()` or drop them — hardcoded English strings can't ship in a UI with 23 locales.

**3. The `model/` layer is missing.** Discussion #1716 calls it "the one contract everything else depends on", and this PR splits the files without introducing it. The renderer is still a ~200-line function. I'm fine with that as a first step, but say so in the PR description so the follow-up isn't a surprise.

**4. Duplication left standing.** The overlay logic still mirrors `TripPDF.tsx:709-749`. Not a blocker for this PR, but it's the kind of thing that's much cheaper to fix during a move than after.

And a question that's mine to answer, not yours: this introduces `client/src/features/` as a new top-level convention. I need to decide whether that's how we organise things going forward and write it into CLAUDE.md — I'll come back to you on that, it shouldn't hold you up.

@jubnl's change request is still recorded against a commit that has since been overwritten; I'll ask him to re-check or clear it.

Could you get back to me within **3 days**? If nothing lands by **6 August** I'll close this to keep the queue current — reopening is always fine.

SimMesg20:
> The existing `JourneyBookPDF.test.tsx` is untouched and still passing, which is the evidence I wanted that this is a move rather than a rewrite. Thanks for keeping it that way.
> 
> Before I merge it, a few things:
> 
> **1. `const document` shadows the global.** `JourneyBookPDF.tsx:11` — inside a file that also deals with the real DOM `document`, that's a trap waiting for the next person. Please rename it (`pdfDoc` or similar).
> 
> **2. Untranslated labels.** `showJourneyExportPreview.ts:6-8` hardcodes `saveLabel`, `closeLabel` and `pagesLabel`. Either route them through `t()` or drop them — hardcoded English strings can't ship in a UI with 23 locales.
> 
> **3. The `model/` layer is missing.** Discussion #1716 calls it "the one contract everything else depends on", and this PR splits the files without introducing it. The renderer is still a ~200-line function. I'm fine with that as a first step, but say so in the PR description so the follow-up isn't a surprise.
> 
> **4. Duplication left standing.** The overlay logic still mirrors `TripPDF.tsx:709-749`. Not a blocker for this PR, but it's the kind of thing that's much cheaper to fix during a move than after.
> 
> And a question that's mine to answer, not yours: this introduces `client/src/features/` as a new top-level convention. I need to decide whether that's how we organise things going forward and write it into CLAUDE.md — I'll come back to you on that, it shouldn't hold you up.
> 
> @jubnl's change request is still recorded against a commit that has since been overwritten; I'll ask him to re-check or clear it.
> 
> Could you get back to me within **3 days**? If nothing lands by **6 August** I'll close this to keep the queue current — reopening is always fine.

I'll see if I can get your feedback in tomorrow if I have the time, otherwise I think the day after I can get it done.

The model structure aswell as a few other folder don't exist yet because the actual code needed for it comes in a PR after that, I can add the folder structure though if that's preferred. Happened do to me focusing on keeping it a refactor as much as possible.

SimMesg20:
@mauriceboe I tried applying your feedback, i am a bit unsure about it as it now goes a bit further then a pure refactor.
I commited it as WIP for now and you can take a look at it, if its to much or you want me to go into another direction thats ok too

mauriceboe:
I am going to close this one, and I want to be clear that it is not a judgement on the work you put in. You asked for a direction decision, the answer took too long, and in the meantime the branch has drifted 160+ commits behind `dev`. Reworking it from here is more painful than starting clean, so let me give you the direction now and what a fresh attempt should look like.

### The decision you asked for

No `client/src/features/`. I do not want a second top level structure next to `components/`, `hooks/`, `services/` and `store/` while the rest of the app uses the existing one. If this comes back, the modules belong in `client/src/components/PDF/` (the render and preview parts) and `client/src/services/` (the snapshot builder), matching what the tree already does.

### What I would actually like to see

Cut it into two, and the first one is easy to say yes to:

**PR 1: the overlay.** `TripPDF` and `JourneyBookPDF` both build the same preview overlay by hand. Lifting that into one module in `client/src/components/PDF/` is a clear win, it is small, and it stands on its own. I would merge that on sight.

**PR 2: the journey book itself.** The split into snapshot, render and preview is the right idea, it just has to land complete:

- **Translations all the way down.** Right now `t` stops at the entry point and the renderer still emits English literals. Every user visible string has to go through `t`, and any new key has to exist in all 23 files under `shared/src/i18n/`, or `node shared/scripts/i18n-parity.mjs --strict` fails the build. Pass `locale` down too and use it for date formatting instead of the hardcoded format.
- **Run `cd client && npm run theme:lint` before you push.** The moved code carries about 35 raw colour and size violations with it. It does not run in CI, so it is on us to run it. If a generated print document genuinely has to sit outside the token system, the honest fix is an entry in the `EXEMPT` list in `client/scripts/theme-lint.mjs` with a one line reason, not silence.
- **One `esc()`.** It exists twice after the split.
- No empty scaffolding directories.
- Start from current `dev`.

One thing worth saying plainly, because the title suggests otherwise: this splits the source file, not the output. Users still get exactly one PDF through the browser print dialog. If what you actually want is a real multi file export (per chapter, or a ZIP), that is a different and more interesting feature, and worth its own issue to agree on the shape first.

Sorry for the long silence on the direction question. If you open the overlay PR I will look at it the same week.

SimMesg20:
@mauriceboe Thanks for the detailed reply, ill probably do the first PR once i have time, but will hold off on the second one for a bit, since i am a bit busy atm, ill rethink what structure (of splitting the file) makes sense and how it can fit the existing folder structure better.

As for your last mention, it was not my intend to split the users actual export file into multiple, although that would be a good idea at a later point esp with the current limit of pictures.
I will see what i can do in the next few weeks

Issue #1453, opened 2026-07-06
[BUG] Subscribed Calendar Events Not Respecting Times Zones


stevenhaskell:
### Pre-flight checklist

- [x] I have searched [existing issues](https://github.com/mauriceboe/TREK/issues) and this bug has not been reported yet
- [x] I am running the latest available version of TREK
- [x] I have read the [Troubleshooting guide](https://github.com/mauriceboe/TREK/wiki/Troubleshooting) and my issue is not covered there

### TREK version

v3.2.1

### Describe the bug

I just enabled the calendar subscription for my upcoming trip. In Trek, it shows the correct time zones for the events. When you add the subscribed calendar, it's showing the event on my local time zone. It should be using the time zone shown in Trek.

### Steps to reproduce

1. Add an event/transportation to a trip with a time zone that is not your current.
2. Subscribe to the trip's calendar in IOS or Google Calendar.
3. Locate the event and it will be showing in your local time zone, not the time zone listed in Trek.

### Expected behavior

The event in the subscribed calendar should be using time zone shown in Trek.

### Deployment method

Docker Compose

### Host OS

Synology DSM 7.3.2-86009 Update 3

### Accessing TREK from

Desktop browser

### Browser (if applicable)

Version 149.0.7827.201 (Official Build) (arm64)

### Relevant logs or error output

```shell
```

### Screenshots

![Image](https://github.com/user-attachments/assets/fa7377ca-5b02-43c8-98f1-c3943a2edb21) ![Image](https://github.com/user-attachments/assets/5fc37f89-2e2a-43db-84cb-ca98cc074830)

stevenhaskell:
@mauriceboe Just saw that this was reopened and closed again. Looks like there is still an issue with the time zones, see attached image. 

![image](https://github.com/user-attachments/assets/fa555f3f-dbf5-4c99-b7bb-0bc198544b7d)

On the left is the event generated by TREK, on the right a manual one I created. The generated one is only using the departure time zone for start/end and not respecting the timezone for the end time.

PR #1721, opened 2026-07-27
feat(feed): TripIt-style hotel and car-rental events in the ICS feed, and fix floating flight times

michael-bohr:
## Description

Brings the ICS calendar feed closer to how TripIt models bookings, so a subscribed calendar stays readable.

- **Hotels** emit a pair of short `TRANSP:TRANSPARENT` marker events, `Check-in: <hotel>` and `Check-out: <hotel>`, instead of one all-day event on the check-in date. Dates/times resolve from the linked `day_accommodations` row (day dates plus `check_in`/`check_out`, tolerating `"15:00"` from the UI and full ISO from booking import), then `metadata.check_in_time`/`check_out_time`, then `reservation_time`/`reservation_end_time`, defaulting to 15:00/11:00. A stored check-in window end (`check_in_end`) becomes the check-in event's DTEND.
  - This also fixes hotels created by **booking import never appearing in the feed at all** — they carry no `reservation_time`, so the old code skipped them silently.
- **Car rentals** emit `Pick up:` / `Drop off:` transparent point events (matching TripIt) instead of one opaque event spanning the whole rental, which marked the subscriber busy for days. Sides resolve by endpoint role, so a surviving return endpoint can't be mistaken for the pickup when import dropped the other side (lat/lng are NOT NULL, so a failed geocode drops the row); each side falls back to `reservation_time`/`reservation_end_time`.
- **Free/busy**: the whole-trip banner and the all-day day-summary events are now `TRANSP:TRANSPARENT`, so subscribing to a month-long trip no longer blocks a month of availability.
- **Timezones**: transports that carry both `reservation_time` and endpoints took the `reservation_time` branch, whose zone comes from the linked place — flights have none — so they still rendered as floating times after #1453. Endpoints are now consulted first, since they hold per-side IANA zones. A SEA→HND flight subscribed from another zone was showing as a ~26 hour block.
- **Invalid events**: a `DTEND` that isn't after its `DTSTART` is dropped when both ends share a zone (or are both floating); cross-zone pairs are left alone because their wall clocks aren't comparable. A time-only `reservation_end_time` earlier than the start (a bar closing at 01:00) now rolls to the next day instead. Real overnight-arrival data was producing negative-duration VEVENTs.
- **Robustness**: one reservation with corrupt `metadata` JSON used to throw and take down the entire feed (and drop the trip from the all-trips feed); it now renders without its metadata fields. Non-string metadata values are ignored rather than thrown on.

UIDs stay deterministic: they're derived from row ids only, and the split bookings add a suffix to the reservation id (`trek-res-42-checkin@trek`, `-pickup@trek`, …), so they can't collide with the single-event `trek-res-42@trek`. Nothing in the feed is keyed off an accommodation id.

Discussed and approved in `#github-pr` on Discord. Follows on from the timezone work in #1453.

## Type of Change
- [x] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update

## Checklist

## Test plan

- 17 new unit tests in `tripService.test.ts` (`TRIP-SVC-021`–`037`): accommodation-sourced check-in/out windows with TZID from the accommodation's place; default, metadata and `reservation_end_time` clocks; the `check_in_end` window; import-shaped hotels (ISO datetimes, no `reservation_time`); legacy bare-date hotels; car endpoints with TZID; role-based car sides and per-side fallback; endpoint priority over `reservation_time`; the same-zone DTEND guard; the overnight bare-time roll; corrupt/non-string metadata; a dated-but-timeless endpoint not downgrading a timed reservation; TRANSP on banner and day events; and UID uniqueness + stability across two exports.
- `TRIP-SVC-004` now uses a non-hotel type — it still covers the generic date-only path, and hotels have dedicated tests.
- The UID test was sanity-checked by making a UID non-deterministic to confirm it fails, rather than trusting a green assertion.
- Full server suite green. Verified end-to-end against a real 27-day, 8-hotel, 7-flight, 2-rental trip on a test instance: 16 window events, correct per-side flight zones, zero invalid same-zone DTENDs.

## Note

Left out deliberately: there's no migration that backfilled hotel reservations for accommodations created before `createAccommodation` started auto-creating one, so in older databases an accommodation with no linked reservation would still be missing from the feed. Covering those means emitting accommodation-keyed events (`trek-acc-{id}-…`); happy to do that as a follow-up if wanted.

### DIFF
--- a/server/src/services/reservationService.ts
+++ b/server/src/services/reservationService.ts
@@ -222,6 +222,14 @@ function saveEndpoints(reservationId: number, endpoints: EndpointInput[]): void 
   tx(reservationId, endpoints);
 }
 
+// accommodation_id is a TEXT column; the integer FK reads back as a numeric
+// string (e.g. "14.0"). Normalize to an int so consumers can use it as a key.
+export function normalizeAccommodationId(value: unknown): number | null {
+  if (value == null) return null;
+  const n = Math.trunc(Number(value));
+  return Number.isFinite(n) ? n : null;
+}
+
 export function listReservations(tripId: string | number) {
   const reservations = db.prepare(`
     SELECT r.*, d.day_number, p.name as place_name, r.assignment_id,
@@ -256,9 +264,7 @@ export function listReservations(tripId: string | number) {
     r.day_positions = posMap.get(r.id) || null;
     r.endpoints = endpointsMap.get(r.id) || [];
     r.travelers = travelersMap.get(r.id) || [];
-    // accommodation_id is a TEXT column; the integer FK reads back as a numeric
-    // string (e.g. "14.0"). Normalize to an int so clients can parse it.
-    r.accommodation_id = r.accommodation_id == null ? null : Math.trunc(Number(r.accommodation_id));
+    r.accommodation_id = normalizeAccommodationId(r.accommodation_id);
   }
 
   return reservations;
--- a/server/src/services/tripService.ts
+++ b/server/src/services/tripService.ts
@@ -9,7 +9,7 @@ import { Trip, User } from '../types';
 import { listDays, listAccommodations, addDays, resyncAccommodationDays, restampReservationDates } from './dayService';
 import { listBudgetItems, removeUserFromBudgetItems } from './budgetService';
 import { listItems as listPackingItems } from '../nest/packing/packing.bridge';
-import { listReservations, loadEndpointsByTrip, resyncReservationDays } from './reservationService';
+import { listReservations, loadEndpointsByTrip, normalizeAccommodationId, resyncReservationDays } from './reservationService';
 import { listNotes as listCollabNotes } from '../nest/collab/collab.bridge';
 import { shiftOwnerEntriesForTripWindow } from './vacayService';
 import { resolveTimeZone } from './timezoneService';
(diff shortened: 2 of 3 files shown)


mauriceboe (changes requested):
This is the most valuable PR in my current queue, and I want it in. A 27-day trip showing up as one solid "busy" block makes the ICS feed useless for anyone who actually shares their calendar, and the floating-time regression on flights after #1453 is real.

The problem is where it's built, not what it does.

**The files this PR changes no longer exist on `dev`.** `server/src/services/tripService.ts`, `server/src/services/reservationService.ts` and `server/tests/unit/services/tripService.test.ts` were removed when the trips domain moved to NestJS. GitHub reports the conflict as modify/delete across the whole diff, so there is no rebase that saves it — it needs porting by hand.

The good news: it's mechanical. The code on `dev` is the same logic, just relocated.

- Feed building moves to `server/src/nest/trips/trips.service.ts` (the ICS section starts around `:918`)
- SQL goes through the injected `this.db`, not the imported `db` singleton
- The `day_accommodations` lookup belongs in `days.service.ts` next to `listAccommodations`, rather than as a second query inside trips
- Tests move to `server/tests/unit/nest/trips.service.test.ts` — note the IDs `TRIP-SVC-021/024/025/026/035/036/037` are already taken there, so yours need renumbering. `TRIP-SVC-004` (`:419`) also asserts `hotel` where your change produces `other`, and `tests/helpers/factories.ts:434` needs `check_in_end`.

`server/src/nest/README.md` has the full recipe.

One behavioural thing I'd like changed: the precedence should be `acc.check_in` **before** `meta.check_in_time`. `days.service.ts:609-611` never deletes metadata keys, so reading metadata first means the feed can show a check-in time the user already deleted.

Two things I'll own, not you: `TRANSP:TRANSPARENT` on the trip banner, and the fact that every existing hotel and car-rental event changes its UID — subscribers will get them as new events and lose their reminders. That goes in the release notes.

Could you let me know within **3 days** whether you want to do the port? If I hear nothing by **6 August** I'll close this so the queue stays honest — but please do reopen, I'd rather have this feature than not.

michael-bohr:
Apologies for the silence, I was on leave and have just caught up. Thanks for the detailed review, and I'd still very much like to get the remaining piece of this in.

I went through current `dev` before starting the port, and most of this has landed in the meantime, so the scope is much smaller than when you wrote that review.

**Already on `dev`, so nothing for me to port:**

- Hotels. `df3edac7` (#1915) added the `stays` query and the `marker` closure in `calendar.service.ts` (~`:417-458`), so a stay now emits the all-day range **and** timed check-in / check-out events, using `check_in_end` as the check-in window's DTEND. `CAL-025`/`026`/`027` cover it. That is the hotel half of this PR, window-end included.
- The flight floating-time regression. `calendar.service.ts` now consults a timed departure endpoint before `reservation_time` for transports, citing #1453, which is the same fix as my `a724b817`.

Also worth flagging for the recipe: feed building has moved out of `server/src/nest/trips/trips.service.ts` (which no longer contains any ICS code) into `server/src/nest/calendar/calendar.service.ts`, with `feeds.service.ts` as the wrapper.

**What's actually left:**

1. **Car rentals.** Nothing emits pickup / drop-off events yet, so a rental still shows as one block. This is the piece I'd like to port: derive the two sides from the reservation's endpoints by `role` (`from` = pickup, `to` = drop-off), falling back to positional first/last, then to `reservation_time` / `reservation_end_time` for imports that only carry the rental window.

2. **A metadata fallback, if you want one.** Your note asked for `acc.check_in` before `meta.check_in_time`, but the merged `marker` reads the stay values only and never touches metadata, so there's no precedence left to invert. If you'd like metadata as a *fallback* for bookings whose clock only lives there, I'd add it strictly below the stay values, which is where your note was pointing anyway. Happy to leave it out entirely if you'd rather keep metadata out of the feed.

`TRANSP:TRANSPARENT` on the trip banner still looks open on your side.

Given how little is left and that the original diff is modify/delete against three files that no longer exist, I'd suggest leaving this PR closed and opening a fresh, small one against `calendar.service.ts` for the car-rental windows. Just let me know on (2) and I'll get it up.

michael-bohr:
Opened #1919 with the remaining piece, the car-rental pickup and drop-off events, as a fresh change against `calendar.service.ts` rather than a rebase of this diff.

It leaves the metadata fallback out pending your answer above, and doesn't touch anything #1915 shipped. Suggest leaving this one closed.

PR #1732, opened 2026-07-28
Add "use my current location" button to journey entry editor

RenzoBeux:
Adds a crosshair button inside the journey entry editor's location field that fills in the device's current position. Mostly useful on mobile, where you're usually writing the entry from the place itself.

How it works:
- New one-shot `getCurrentPositionOnce()` helper in `useGeolocation.ts` (the existing hook is watch-only, meant for map follow mode)
- On click it sets the coordinates immediately, then reverse geocodes through the existing `/api/maps/reverse` to fill in a place name — if that returns nothing, the field keeps the plain coordinates
- Permission denied / timeout / non-HTTPS origins each get their own error toast
- No server changes

The button is visible on desktop too (unlike the map FABs): since the result lands in an editable field the user can review and correct, an IP-based fix isn't misleading here.

i18n keys were added to all 23 locales; the non-English strings are machine translated, happy to fix any of them.

Tests: unit tests for the helper, component tests for the fill/fallback/denied flows.


mauriceboe (changes requested):
The helper itself is good — `getCurrentPositionOnce()` with the insecure-context guard is the right shape, there's no server surface, and the tests are better than average for a UI addition.

But the PR argues it's "mostly useful on mobile", and on mobile the button never appears: `client/src/pages/JourneyDetailPage.tsx:31` switches to `MJourneyEntrySheet` below 768px, and that component doesn't render it.

So one of two things:

**Either** wire it into `client/src/mobile/screens/journey/MJourneyEntrySheet.tsx` as well — note `:86-111` already does its own raw `navigator.geolocation` call, and `:106` surfaces the untranslated `error.message` to the user. Switching it to your helper fixes both and means there's one implementation instead of a fourth variant.

**Or** correct the PR description, and I'll merge it as a desktop-only convenience.

I'd prefer the first, but I'll take either.

One smaller thing: the race guard in `JourneyDetailPageEntryEditor.tsx:222` deserves a test; it's the part most likely to break silently later.

Could you get back to me within **3 days**? If I hear nothing by **6 August** I'll close this to keep the queue honest — reopening later is always fine.

RenzoBeux:
Went with the first option. `MJourneyEntrySheet` now has the same crosshair button in its location field, and I switched its quick-capture auto-locate over to `getCurrentPositionOnce()`, so the raw `navigator.geolocation` call and the untranslated `error.message` at :106 are gone. The error code to message mapping moved into a small shared helper (`geoOnceErrorKey` in `JourneyDetailPage.helpers.ts`) so both editors surface the same translated strings. No new i18n keys were needed, the mobile sheet reuses the existing `journey.editor.*` ones.

Also added the race guard test you asked for: the reverse geocode is held in flight while a search result gets picked, then released. The test fails if the guard ever stops protecting the pick.

PR #1734, opened 2026-07-29
Raise the frontend test coverage back over the gate

mauriceboe:
The client coverage had drifted to 47.98% statements / 50.66% lines while CI stayed
green. The reason is in `client/vitest.config.ts`: it never had `coverage.thresholds`,
so the Client Tests job ran `test:coverage`, uploaded the report as an artifact and
passed regardless of the number. Only the server is gated (`server/vitest.config.ts`,
80% over `src/nest/**`).

This adds the missing tests and the missing gate.

## Coverage

|            | before | after  |
|------------|--------|--------|
| Statements | 47.98% | 96.76% |
| Branches   | 43.18% | 89.88% |
| Functions  | 45.54% | 96.43% |
| Lines      | 50.66% | 98.11% |

The untested paths were the mobile rewrite and the 4.0.0 show — neither had ever been
covered:

```
1813  1%  mobile/screens/trip/tabs      1121  8%  mobile/screens/settings
1606  1%  mobile/screens/trip/sheets     930  6%  components/FourZero
1493  5%  mobile/screens/admin           621 19%  components/Collections
```

Both are at or near 100% now, including `noFearAudio.ts` (376 lines of Web Audio,
covered against a stand-in AudioContext) and `noFearScene.ts` (canvas engine).

## What is in here

- Tests only. No production file is touched — `git diff` over `client/src/**` shows
  nothing but `*.test.tsx` additions.
- Three shared fixtures under `client/tests/helpers/`: `mobileTrip.ts`
  (`buildPlanner`/`buildShell`, since every screen under `src/mobile/screens/trip`
  takes the same two props), `mobileAdmin.ts` and `atlas.ts`.
- `coverage.thresholds` in `client/vitest.config.ts`, set a few points below the
  current run so an ordinary PR does not trip them.

## Findings

Writing the tests surfaced 214 things in the production code — real defects, dead
code and a few a11y problems. They are listed in the first comment below. The tests
here pin the current behaviour, including the wrong parts; the follow-up PR fixes
them and moves the affected assertions with it.

The ones worth reading first:

- `MPackingListTab.applyTemplate` calls `packingApi.applyTemplate(tripId, templateId)`
  without `visibility`, so applying a template while the list toggle sits on
  "personal" drops the items into the common list where the user cannot see them.
  The desktop panel passes it.
- `useMPlanTimeline.applyMergedOrder` writes transport positions and leg metadata to
  the store optimistically and does not roll them back when a later await rejects.
[...]

PR #1737, opened 2026-07-29
Work through the findings from the coverage pass

mauriceboe:
Follow-up to #1734. Writing those tests turned up 214 things in the production
code; they are listed there as F001–F214. This works through all of them.

|                    |     |
|--------------------|-----|
| corrected          | 150 |
| dead code removed  |  64 |
| left alone         |   0 |

The tests in #1734 pinned the behaviour as it was, including the wrong parts.
Every fix here moves its pinning test with it, so neither PR is green on its own
premise and then red once both are on `dev`.

## The ones worth reading

- **Packing templates landed in the wrong list.** `MPackingListTab.applyTemplate`
  called `packingApi.applyTemplate(tripId, templateId)` without `visibility`, so
  applying a template while the toggle sat on "personal" dropped the items into
  the common list, where the user could not see them.
- **A failed reorder left the UI lying.** `useMPlanTimeline.applyMergedOrder`
  wrote transport positions and leg metadata to the store optimistically and did
  not roll them back when a later await rejected.
- **"My" to-dos showed other people's.** `filterTodoItems('my', …)` compared
  `assigned_user_id === currentUserId` with a null user id, so every *unassigned*
  item matched. `todoCounts` right below it already guarded for exactly that, so
  the badge and the list disagreed. Same line in `useTodoList.ts`.
- **A booking with a date but no times could not be saved** once an end date was
  entered — `isEndBeforeStart` compared `${end}T00:00` against a start it had
  padded the same way.
- **Downloads aborted in Firefox and Safari.** `MExportSheet` and `MCostsTab`
  revoked the object URL in the same tick as `a.click()`, from an anchor that was
  never in the document.
- **A note attachment came back from the dead.** Removing one dropped the chip
  before the request settled and swallowed the rejection, so a failed DELETE left
  the file gone on screen and present on the server.

## Dead code

A little under a third of the findings were code no test could reach: functions
returned from a hook, destructured in the component and never called; guards
whose false arm is unreachable; error branches behind a condition a disabled
button already enforces. The largest was `ReservationOverlay.tsx`, where
`buildStatsHtml` and the whole `StatsLabel` component — 164 lines — were never
rendered and `showStats` was never read. Its Mapbox counterpart documents the
removal in a comment; the Leaflet side was never cleaned up.

Those are deleted rather than covered.

## Notes

- New i18n key `places.imageRemoveError`, in all 23 locales — removing an image
  used to report "Could not upload image".
- The mobile transport sheet now writes the day from the filtered stations
  (F019/F020). The desktop `TransportModal.tsx` still has the same bug; the
  payload contract is unchanged, only the values now match the endpoints that are
  actually written. Pulling desktop along is a separate call.
- Also drops `client/coverage-agentF/` and `client/coverage-jsheet/`, two scratch
  files that slipped into #1734, and adds the ignore rule that keeps them out.

## Coverage

Removing the dead code moved it up rather than down:

|            | #1734  | here   |
|------------|--------|--------|
| Statements | 96.76% | 97.55% |
| Branches   | 89.88% | 90.61% |
| Functions  | 96.43% | 96.92% |
| Lines      | 98.11% | 98.82% |

Suite green, typecheck clean, eslint 0 errors, i18n parity OK.


mauriceboe:
## Outcome per finding

IDs match the list on #1734. 150 corrected, 64 removed as dead code, none left open.

### Corrected

**F002** NoFearShow.tsx: switches the resize listener to a 150ms debounce (`onResize` -> setTimeout(fit)) and clears the timer in the cleanup; a window drag now bakes the static layers only once.

**F003** NoFearShow.tsx: `fit()` sets `assemblyRef.current = null` so the particle targets are re-sampled against the moved title rect after a resize.

**F004** noFearAssembly.ts: `init` resets `ready`/`particles` before the getContext early return (the later duplicate `particles = []` assignment is dropped), so a failed re-init goes silent instead of stale.

**F005** listsModel.filterTodoItems and useTodoList: with an unresolved currentUserId, `my` now returns [] instead of all UNASSIGNED rows — matching todoCounts/myCount.

**F006** close() now also resets setColorPickerOpen(false) so the track colour picker starts collapsed on the next place

**F007** handleImageRemove reports the new key places.imageRemoveError instead of places.imageUploadError; key added in all 23 locales

**F008** dayLabel uses day.day_number || dayIndex + 1 so a day_number of 0 falls back to the position instead of 'Day ?'

**F009** Filter state is now tagged ({kind:'smart'|'category'}); category buckets run through the new filterTodoItemsByCategory() instead of through the filter ids

**F010** Removed the dead '|| item.due_date' fallback and replaced it with a real date check so an unparsable due_date shows raw instead of as 'Invalid Date'

**F012** handleVote compares against the requested pollId like handleClosePoll does, instead of against res.poll.id

**F013** HotelConnRow uses the same TravelLine as ConnRow and therefore respects seg.mode/durationText/noteText (#1281) instead of a fixed Footprints+Car

**F014** The connector tap for switching modes now hangs off `editing` instead of `canEdit`, so it is no longer reachable in Go mode

**F015** The row map is driven through `day &&`; both `tl.day!.id` assertions are gone and dayId comes from the narrowed const

**F016** TransitStrip now only filters out WALK legs with a known duration < 60s; a leg without a duration stays in (minutes then empty instead of 0)

**F017** reduce-motion lives in state and subscribes to the MediaQueryList change event (with cleanup); the dead `typeof window` branch is gone

**F018** The central buildPlanner fixture already carries `visibleConnections: [] as number[]`; the redundant local override in MTransportSheet.test.tsx was removed

**F019** firstTrainWp/lastTrainWp come from the filtered trainStations (flight parity); only a train with no selected station at all falls back to the raw rows

**F020** Stopovers without metadata.legs are seeded from endpoint.local_date when editing (stopDayId) instead of losing the day with an empty arrDayId/depDayId

**F021** The day label uses d.day_number || i + 1; the unreachable '?' branch is removed

**F023** MPlaceEditSheet: hasTimeError no longer depends on sheetPlace, only on the time values themselves

**F024** MCollabNotes: handleDeleteFile now returns true/false; the chip only disappears after a confirmed DELETE, and on success the attachment is additionally removed from the notes state

**F025** MCollabNotes: the eyebrow above the textarea uses the new key collab.notes.content instead of the placeholder key (23 locales updated)

**F026** MCollabNotes: attachment chips get aria-label collab.notes.removeFile with the file name (new in 23 locales), only the card delete button keeps collab.notes.delete

**F027** MCollabChat: hasReply only depends on reply_text so a reply to a deleted message no longer renders an empty quote box

**F028** MCollabChat: MessageActionsPopover now also clamps top against window.innerHeight (height including the optional delete row), with the dimensions as constants

**F029** ImpFileStep: a mixed import counts GPX and KML separately and toasts each format's own key; the undo label for mixed imports is the new undo.importFiles (23 locales)

**F030** ImpFileStep: skipped for KML/KMZ falls back to result.skipped when there is no summary, so importAllSkipped warns in that case too

**F032** MFileLightbox: a click on the empty media area now stops propagation, so onClose fires only once

**F033** MFilesTab: uploadFiles catches per file, keeps uploading the rest of the batch and reports partial success (success toast with the real count plus an error toast)

**F035** MFileTrashSheet: an emptying state disables the empty button and all row actions (restore/delete) while the empty request runs

**F036** MExportSheet.downloadIcs: the anchor is appended to the DOM, revokeObjectURL + remove happen only after 100 ms (same pattern as utils/fileDownload.ts)

**F037** MExportSheet.downloadIcs: icsBusy guard analogous to pdfBusy, the row shows common.loading in the meantime

**F038** MNoteSheet: the seed effect now depends on payload.note.id + payload.dayId instead of on the payload object identity, so an inline-built payload no longer overwrites what is being typed

**F039** MCostSheet.save(): a selected payer is sent even without participants, otherwise the server derives total_price as 0 from the empty payer sum

**F040** MPackingListTab.applyTemplate passes the current view (common/personal) as visibility to packingApi.applyTemplate

**F041** MPackItemSheet: the prefill effect depends on [open, item.id] instead of on the item object, so a WebSocket refresh no longer overwrites the draft

**F043** MPlacesBrowser: bulk category and bulk delete are awaited; select mode is only left after a successful mutation, otherwise the selection stays around for a retry

**F044** MPlacesBrowser.allSelected compares the ids (filtered.every(p => selectedIds.has(p.id))) instead of just the sizes

**F045** The MTripShell day chip falls back to the list position (idx + 1) instead of to 'Day 0' when day_number is missing

**F046** MReservationSheet.isEndBeforeStart compares only the dates when there is no time, so an all-day booking with an end date on the start day stays saveable

**F047** MAccommodationSheet.back() falls back to editing.start_day_id when the payload carries no dayId while editing

**F048** MAccommodationSheet.save(): the try block only wraps the API write; loadAccommodations() and back() run afterwards, so an error there is no longer reported as a failed save

**F050** useMPlanTimeline.applyMergedOrder catches up on refreshDays + loadReservations in the catch, so the optimistic transport/leg positions do not stick around after a failed reorder

**F051** New export downloadBlob(blob, filename) in utils/fileDownload.ts (uses the correct triggerAnchorDownload: append -> click -> revoke after 100ms + a.remove()); MCostsTab.handleExportCsv and CostsPanel.handleExportCsv no longer build the anchor themselves.

**F052** MCostsTab settle-up header: the role=button div with a nested button is replaced by two sibling buttons (toggle with chevron + 'Add payment'), stopPropagation is gone

**F053** New ConfirmationCode component in tabChrome.tsx: with blur active the code box is a real button (aria-pressed) instead of a div onClick; for that, the card bodies in MTransportsTab/MBookingsTab are split around the code into two sibling buttons so nothing interactive is nested

**F054** MAdminGitHubPanel: setError(null) on every successful load; the error card now only replaces the timeline when nothing is loaded yet, otherwise the timeline stays and the error shows inline above a still-clickable Load more (retry path).

**F055** Pagination now walks further pages (bounded by MAX_PAGES_PER_LOAD=5) until at least one release survives the prerelease filter, so an all-prerelease page can no longer leave an empty timeline behind a Load more button.

**F057** MAdminNotifyMatrix: event labels look the key up first (labelKey ? t(labelKey) : eventType) instead of calling t(undefined); channel labels always have a key, so the unreachable `|| ch` fallback is gone.

**F058** MAdminDevNotificationsPanel: Btn hoisted to module scope with an explicit `sending` prop, TripSelector/UserSelector inlined at their single call sites (no remount, selects keep focus/selection), and `catch (err: any)` replaced by `unknown` + instanceof.

**F059** adminApi.llmLocalPull now wraps only JSON.parse in try/catch and calls onProgress outside it (reader cancelled in finally), so a streamed Ollama error frame aborts the pull instead of ending in toast.success('Model pulled') — fixes mobile and desktop AddonManager alike.

**F060** Category and item delete failures now toast admin.toast.deleteError ('Failed to delete') instead of 'Failed to delete template'; the key already exists in all 23 locales, so no i18n additions were needed.

**F061** The toast now builds the server detail text as a suffix (`${msg}${detail}`) — no space before the colon, no trailing space when there is no error

**F062** settings_get only writes providerConnected when the provider has no status_get route — the status route decides the badge, the race is gone

**F063** onOfflinePrefsChange now returns a real void unsubscribe (block instead of expression body)

**F064** NoFearShow.tsx: the ghost id is captured in `const ghostId` before the `lastCue` reassignment, so the 1400ms timer actually clears the ghost row.

**F065** The key 'common.success' now exists in all 23 locales (created on this branch, my duplicate line removed again); FE-MOB-SETPLG-013 expects 'Success' and passes — the source needed no change

**F066** reloadClients() re-reads the client list after create and after rotate instead of appending optimistically

**F071** New key settings.avatarRemoveError in all 23 locales; handleAvatarRemove no longer reports 'Upload failed'

**F072** The registry detail preview now carries the UI-facing capabilities (widget.slot, tripPage.replaces) from the server, and PluginDetailSheet passes manifest.capabilities to deriveCaps — the 'Replaces planner tabs' chip is visible before installing (desktop panel fixed the same way).

**F074** onDragLeave is symmetric with onDragEnter: a leave without a matching enter returns early instead of driving dragDepth negative.

**F076** loadAutoSettings only calls setAutoSettings when the 200 actually carries a settings block (same guard on the save echo), so a malformed response no longer blanks the schedule state and crashes the next render.

**F077** types.ts getApiErrorMessage returns the localized fallback for HTTP errors whose body has no `error` field instead of axios' 'Request failed with status code …'; MAdminDefaultUserSettings now routes both catches through it so common.error is reachable.

**F078** Map-preview place fixture uses `new Date().toISOString()` instead of `Date()`, matching the ISO timestamp the Place shape expects.

**F079** MCollPlaceSheet, MCollAddSheet and MCollEditSheet now import getApiErrorMessage from utils/apiError — no more raw 'Network Error' in the place sheet, all collections sheets behave the same

**F080** pick() catches a rejected onPick (catch + finally) instead of letting it escape as an unhandled rejection

**F081** Label delete goes through remove() with a busy guard/await and the button is disabled in the meantime — serialized like create/update

**F083** The target effect depends on lists again (no eslint-disable): defaulting to the only list also applies when the lists load only after opening; a pick already made is kept, targetId is reset on close

**F084** close() hands a list that was created but only failed at the cover step to the parent via onCreated — cancelling no longer leaves an invisible list behind; the createdId retry is preserved

**F085** The How-to-Update sheet links https://github.com/liketrek/TREK/wiki/Updating; the docker pull line stays on mauriceboe/trek because that is still the published image (docker.yml, README) — only the GitHub org moved.

**F086** MAdminMcpTokensPanel keeps t/toast in a ref that is refreshed on every render, so the once-only loader toasts its load errors in the locale that is current when the request fails.

**F087** The audit log offset moved from state into a ref (never rendered), so loadMore no longer captures a spent page number and its useCallback has no deps. Note the duplicate-append scenario itself was already blocked: MAdminButton sets disabled={disabled || busy}, so the Load more button IS disabled while the request runs.

**F088** 'common.success' newly created in all 23 locales (shared/src/i18n/<loc>/common.ts) so PluginSettingsTab:211 and MSettingsPlugins:239 render 'Success' instead of the raw key

**F090** MobileEntryCard resolves the photo id through one helper (photo_id ?? id) for both the public and the authenticated path; no more /api/photos/undefined/thumbnail

**F091** JourneyMapGL got a dedicated basemap-language effect ([mapLang, isMapLibre, glStyle]); the load handler reads mapLangRef so a language switch repins labels without rebuilding the map

**F092** initial-focus effect now also keyed on mapEntries.length and calls the current syncMapToCarousel (and the current index) through refs, so late-arriving markers still get focused

**F096** coordPaste is now also on the longitude NumericInput, so a pasted coordinate pair is split across both fields

**F097** ListEditorModal now imports getApiErrorMessage from ../../utils/apiError instead of from ../../types, so a plain Axios error no longer ends up as raw text in the toast

**F098** The N+1 detail fetches now run in loadSavedPlaces() in batches of four instead of as one Promise.all salvo and are deduplicated by id; no new server endpoint, because the new backend surface (Zod contract + Nest module + tests) lies outside this frontend package

**F099** handleSetRole bails out like handleCancel/handleRemove while settingRoleId is set — simultaneous role changes on two rows are thereby serialized

**F100** Subtitle guard is now `r.address && r.name && r.name !== r.address` in LocationSelect.tsx and AddressInput.tsx, so a nameless geocoder hit prints its address once instead of twice

**F102** dayPickerPlace state typed as useState<Place | null>(null) instead of inferring `null`

**F103** AirportSelect catch is `catch (err)` with a narrowed `(err as { name?: string } | null)?.name` abort check — no `any`

**F104** Optional collection props of MapViewGL get module-scope constants (NO_PLACES/NO_ROUTE_VIAS/NO_ROUTE_SEGMENTS/NO_DAY_ORDER/NO_RESERVATIONS/NO_CONNECTION_IDS/NO_POIS) instead of inline `= []`/`= {}`, so the reconcile effects do not rebuild all markers on every render.

**F105** The photo cache key in MapViewGL runs through the new helper photoCacheKey(); without a provider id and without coordinates it yields '' instead of the literal string "null,null", so the `if (!cacheKey) continue` guard takes effect.

**F107** CollectionPlaceDetail imports getApiErrorMessage from ../../utils/apiError; a rejected upload now shows 'Could not upload image' instead of the raw JS message

**F108** New key 'collections.searchLists' in all 23 locales; SaveTripPlacesToListModal uses it instead of 'collections.copyToTripSearch' ('Search trips')

**F109** SaveToCollectionModal locks all rows like MSaveToCollectionSheet (disabled={busyId != null}); no more visibly ineffective click on a second row

**F110** the nested AddTripDialog is wrapped in a stop-propagation container, so clicks inside the Link-Trip dialog no longer close/discard-prompt the settings dialog

**F111** JournalBody styles react-markdown's own <pre> via a pre renderer; the code renderer only returns <code>, ending the nested <pre><pre> markup

**F112** usePoiExplore clears the loading key in the finally block even when the aborted request was not superseded by a new controller; PoiCategoryPill only shows the spinner for active categories.

**F116** MapView uses the same photoCacheKey() helper for the fetch key and the marker lookup; places without coordinates no longer share a "null,null" entry and the previously unreachable guard now takes effect.

**F117** Restored the missing trigger: a Plus button next to the category select sets showNewCategory, so the inline create-category form, handleCreateCategory and the onCategoryCreated prop are reachable again; added i18n key places.newCategory to all 23 locales

**F121** printBackupCodes now interpolates backupCodesText and the date into the popup HTML through escapeHtml from @trek/shared

**F122** handleTestProvider builds the payload only in the POST branch; the GET probe path (test_get/status_get) no longer silently discards typed credentials but is documented as a probe against the stored credentials

**F123** setMapboxStyle(glStyle) runs only after the successful await updateSettings(...), so a failed save does not leave the normalized style behind in the form

**F127** legIcon and MODE_GROUPS no longer use lucide's `Train` (an alias of TramFront): rail renders TrainFront, subway renders TrainFrontTunnel, so rail/subway/tram are three distinct glyphs

**F129** The cause was in adminApi.llmLocalPull (client/src/api/client.ts) and is already fixed on this branch — only the JSON.parse is still swallowed, a throw from onProgress aborts the pull. The guard in AddonManager.tsx:382 is therefore live; FE-ADMIN-ADDON-027 already pins the correct behavior and passes green. No change needed from me.

**F130** Two new keys in all 23 locales ('admin.packingTemplates.deleteCategoryError'/'deleteItemError'); handleDeleteCategory and handleDeleteItem now report the correct object instead of 'Failed to delete template'

**F131** fire() uses getApiErrorMessage(err, 'Failed') instead of err.message — the toast shows the API's error field instead of 'Request failed with status code 500'

**F139** The arrow move now stores its target row (fromType/toType/toId/insertAfter/toLegIndex) instead of a place-id projection, so confirming the time removal replays the move across bookings and notes; the reorderIds branch and state field are gone

**F141** deleteCollection() and leave() now also clear labels and labelFilter, like setActive(null) and the loadCollection catch.

**F142** Rollback plus re-throw for deletePlace, deleteMany (including selectedIds/selectMode), updateLabel, deleteLabel and assignLabels; the callers in useCollections.ts already catch and toast.

**F143** moveToList() remembers the ids that were actually moved, removes only those locally, runs loadAll() and loadCollection() in every case and only throws the error afterwards.

**F144** placeRepo.update() writes trip_id: Number(tripId) into the optimistic row so it is visible to where('trip_id') reads and is picked up by clearTripData().

**F145** cacheFilesForTrip counts separately: downloaded (real writes only) triggers enforceBlobBudget, present feeds filesCachedCount; SyncMeta.filesCachedCount now has a doc comment that pins down the meaning.

**F146** cacheFilesForTrip(tripId, files) gets the trip id from the caller (file-local, non-exported helper) instead of from files[0]?.trip_id with a truthiness guard; no more falsy-id short-circuit, and an empty file list correctly sets the counter to 0.

**F147** deleteCurrentUserDb() goes through switchTo(ANON_DB_NAME) when the DB is already anonymous instead of opening a second connection to the same name.

**F148** The reload in the catch of reorderBudgetItems/reorderBudgetCategories is now itself inside try/catch; the notify toast runs in every case and nothing escapes as an unhandled rejection.

**F149** reorderPackingItems / reorderTodoItems / reorderBudgetItems filter unknown ids BEFORE the reindex so a stale id leaves no gap (0,2,2) in the sort_order sequence.

**F150** deletePlacesMany: non-null assertion removed, the filter now uses the same predicate as the some() guard (a.place?.id == null || !idSet.has(a.place.id)).

**F152** Added ca, gr and vi to RATE_LIMIT_MESSAGES; all 23 app locales are now covered.

**F153** adminApi.llmLocalPull separates !res.ok (status code message) from !res.body; a 200 without a stream now reports 'Pull returned no progress stream' instead of 'Pull failed (200)'.

**F155** The 429 handler excludes arrays (!Array.isArray(data)); an array body is replaced instead of being given an error property.

**F156** The mobile plan drawer's onRouteCalculated gets the else branch (setRoute(null)/setRouteInfo(null)) like the desktop instance

**F157** Changed hiddenPlaceIds.delete(a.place?.id) to `if (a.place?.id) hiddenPlaceIds.delete(a.place.id)`, mirroring the add branch

**F158** Added activeTab to the dependency list of the invalid-tab guard; FE-TP-HOOK-016 now enables the budget addon (otherwise it pinned a tab that does not exist at all), and the new test FE-TP-HOOK-018b covers the immediate re-validation

**F159** Straightened FileManager onUpdate to `() => tripActions.loadFiles(tripId)` (removed the ignored parameters); the hook now carries a comment that onUpdate is only a refresh signal without a payload

**F161** Removed `&& r.manuallyMarked` from the remaining-regions check (desktop parity AtlasPage.tsx:360), pointed the comment at the server-side rationale (#1490)

**F163** Removed the surplus </div> in the country tooltip template; the markup is now balanced

**F164** The always-empty countryCodeToFlag branch is gone; non-alpha-2 codes get a MapPin placeholder in the flag slot, import removed

**F166** toggle() reads and writes the preferences through a prefsRef instead of the render closure — rollback AND the optimistic update are now state-relative. A pure rollback fix was not enough: with two clicks in one React batch the optimistic updated already clobbers the first change. New test FE-ADMMTX-012 (verified red against the old implementation)

**F167** `if (!editingUser) return` at the start of handleSaveUser, as with the other consumers; new test FE-ADMHOOK-036b

**F168** tripDates now builds the keys from the local date parts (getFullYear/getMonth/getDate) instead of via toISOString(); the test checks the literal keys 2026-05-01..03 instead of just the shape

**F169** Stabilized entries with useMemo — syncMapToCard/pickNearestCard keep their identity, the scroll effect no longer re-attaches on every render and does not lose the 150ms settle timer

**F170** journeyStore.clear() also resets notFound: false so the 404 flag does not stick across a store reset.

**F171** New applyCoverUpdate(tripId, coverUrl) in useDashboard patches trips AND archivedTrips; it replaces the exported setTrips (whose only consumer was the cover handler) in DashboardPage and MDashboard. New test FE-HOOK-DASH-026

**F174** CalendarEditor and AddCalendarDraft strip the `|group:` suffix before deriving the country (like VacaySettings.tsx) — group calendars keep their country and region picker

**F176** CollabNotesWebsiteThumbnail.tsx: ogCache is now keyed with `${tripId} ${url}` (linkPreview is trip-scoped), `failed`/`data` are reset when the url changes and an in-flight fetch is discarded on the change.

**F177** FileManagerImageLightbox.tsx: the keydown effect depends on [onClose] (Escape hits the current callback); ThumbImg only mints the one-shot token via IntersectionObserver once it becomes visible, with an immediate fallback when no IO is available.

**F178** uploadFilesResilient resets loadedMap[i] to 0 before every retry and emits progress so the bytes of the aborted attempt do not remain in the total until the retry's first onUploadProgress.

**F179** Moved the paste normalization to BudgetPanel.helpers.ts as `normalizePastedAmount`; InlineEditCell and AddItemRow now both call the same function.

**F183** Identical to F116 — the template literal fallback is gone, the guard is reachable and is covered by a test.

**F184** endpointIcon builds labelHtml once in full (including the style attribute) and interpolates it; the second escapeHtml(label) call is gone and the output is byte-identical.

**F186** the mobile sheet's prev/next preview slots are marked inert + aria-hidden, so their duplicate pager/title/CTA are out of the a11y tree and tab order

**F191** The day header now counts only assignments that have a place (dayPlaceCount), matching the rows actually rendered; FE-PAGE-SHARED-025 expects 3 places instead of 4 and the comment describes the corrected behavior

**F192** guest rename commits once per editing session (ref gate set on Enter/Escape, reset when the editor reopens or the request failed), so Enter + blur no longer fires two identical PUTs

**F196** fileRef is typed useRef<HTMLInputElement>(null) and the cover path handlers (handleCoverSelect/handleCoverChange/uploadCoverNow) carry real File/ChangeEvent types again

**F197** the provider-discovery effect always clears providersLoading and applies its result (the cancelled flag is gone), so leaving the External-photos tab mid-request no longer strands the spinner and blocks every later run

**F198** STATE_APPLIERS['assignment:reordered'] returns {} when orderedIds is missing instead of clearing the day list - consistent with budget:reordered and day:reordered.

**F199** STATE_APPLIERS['budget:reordered'] indexes through the combined array (reordered + remaining) so that sort_order and array position match.

**F200** Deleted client/coverage-agentF/ and client/coverage-jsheet/ (tracked) as well as client/coverage/ (untracked) - the deletion sits unstaged in git status, no git add per the assignment; added coverage-*/ to .gitignore. The coverage-agent-costs-todo named in the finding does not exist in this worktree; coverage-jsheet was tracked instead.

**F201** ConfirmDialog.tsx: `runConfirm` swallows a rejected promise from onConfirm; CollabNotes:529 returns the promise from the arrow instead of swallowing it in the block. The failing DELETE no longer produces an unhandled rejection, and CollabNotes 210-211 are now covered.

**F204** CollabPolls.tsx: the maximum vote count is computed once via reduce as `topCount` (0 for an empty options list) instead of Math.max(...[]) = -Infinity per option; isWinner now checks `count > 0 && count === topCount`.

**F209** CostsPanel.tsx:766: replaced `!isUnfinished` (a function object) with the already computed `!unfinished` — the 'you lent/you borrowed' line now renders, costs.youLent/costs.youBorrowed are no longer dead.

**F210** TodoListPanel.tsx: hoisted `SidebarItem` out of the render body to module level, taking active/compact/onSelect as props — React sees the same component type, so the filter buttons are no longer remounted on every re-render.

**F211** TripPDF.tsx:436: `<html lang="${(loc || 'en').split('-')[0]}">` — an export without a locale no longer throws.

**F212** PluginDetailModal reads permissions/egress/settings through local defaults (?? []) and the local RegistryDetail manifest fields are declared optional — a registry response without these lists no longer breaks the detail modal

**F213** Wrapped the transport-row metadata JSON.parse in try/catch (matching the guarded parse at :655); a second unguarded parse in the same render path (transport row header) had the same defect and is guarded too

**F214** TripPDF.tsx: `assignments` is defaulted to `{}` once in the destructuring of downloadTripPDF (dayCost and the daysHtml loop also read it unguarded); the `|| {}` at :116/:168/:176 that became redundant are gone.

mauriceboe:
### Dead code removed

**F001** noFearScene.ts: removed `if (thin.length >= 4)` — the preceding `r.length < step * 4` guard already guarantees ceil(r.length/step) >= 4.

**F011** The no-op pill for a stored foreign category is now a <span> label instead of a permanently active button with no effect

**F022** PlPlaceSearch: removed the unreachable 'AbortError' arm, the guard now only checks axios' CanceledError (comment explains why)

**F031** ImpFileStep: removed the unreachable 'if (incoming.length === 0) return' in selectFiles

**F034** MFilesTab: removed the dead ': 0' fallback in openRow, setLightboxIndex gets the findIndex directly

**F042** MPackItemSheet.handleSave: removed the unreachable '!trimmedName || saving' guard (identical to submitDisabled)

**F049** Removed unreachable submit guards: MReservationSheet '!form.title.trim() || isSaving' and MAccommodationSheet '!form.place_id || saving'; for that, place_id was typed as number|'' so the contract types fit without the guard

**F056** MAdminNotificationsSection.testAdminWebhook: dropped the `if (!url && … !== '••••••••') return` guard that the disabled Test button makes unreachable.

**F067** Removed `fields ?? []` in both places — after the earlier return, fields is never null

**F068** Removed the unreachable !addPwd guard including its toast in handleAdd (the button is disabled)

**F069** Removed the unreachable !deletePwd guard in handleDelete

**F070** Removed the three !backupCodesText guards in copy/download/printBackupCodes

**F073** ReviewedBadge lost its unreachable non-compact branch and the `compact` prop; both call sites now render the icon badge directly.

**F075** MAdminCategoryManager.handleSave: dropped the `!form.name.trim()` guard (categories.toast.nameRequired) that the disabled Save/Create button makes unreachable.

**F082** Deleted MCollListsDrawer.tsx and its test (no importer in src/). The missing mobile invite acceptance therefore remains open — that would be a feature in MCollections, not a fix

**F089** Removed the unreachable !addPwd/!deletePwd guards in handleAdd/handleDelete (the buttons carry disabled={busy || !pwd}); the i18n key stays because MSettingsAccount still uses it

**F093** dropped firstId/lastId and the unused isFirst/isLast in JourneyDetailPageMapView

**F094** removed the unused useState and ChevronDown imports in MobileEntryView

**F095** removed the unreachable `if (mapRef.current) remove()` guard at the start of the JourneyMap build effect (cleanup already nulls the ref)

**F101** Dropped the hard-coded `isNaverListImportEnabled` flag and the unreachable useEffect that reset listImportProvider; availableListImportProviders is now a plain ['google','naver'] and handleListImport takes listImportProvider directly

**F106** Removed the dead `if (!lpStart) return` branch in the long-press timer; the timer now closes over a local `start` copy of the touch position.

**F113** Removed buildStatsHtml and the entire StatsLabel component from ReservationOverlay.tsx, along with AIRPORT_BADGE_HALF_PX/BADGE_GAP_PX, the unused imports (useEffect/useRef) and the item fields primaryArc/fallback that were computed only for them.

**F114** Removed buildStatsHtml, the statsMarkers field, the 'render' listener and updateStatsRotation from reservationsMapbox.ts; primaryArc was dropped along with them as their only consumer.

**F115** Removed escAttr and the ZoomTracker component from MapView.tsx; the misplaced JSDoc block above escAttr moves to createPlaceIcon, which it describes.

**F118** Removed the accommodations re-hydrate effect in ReservationModal — the init effect already lists accommodations in its deps and seeds the hotel fields from editAcc first

**F119** Dropped the `if (!dateStr) return ''` branch in ReservationModal's formatDate; both call sites already guard on day.date

**F120** Removed the three ntfyToken !== '••••••••' comparisons (the load effect never puts the mask into state) and noted briefly at the test call that token:null makes the server fall back to the stored token

**F124** Deleted the never-rendered InfoChip component and its InfoChipProps interface from DayDetailPanel

**F125** Deleted the unwired handleToggle in ReservationCard plus the now-unused `toggleReservationStatus` store destructure

**F126** Deleted the unused Row component and RowProps interface from PlaceInspector

**F128** Removed the unreachable `if (mode === 'WALK') return Footprints` branch from legIcon (LineBadge only renders for non-WALK legs) and noted the precondition in a comment

**F132** Removed the unreachable non-compact branch of ReviewedBadge along with the compact prop; both call sites now use <ReviewedBadge t={t} />

**F133** Removed the Leaflet branch in styleForProvider and narrowed the parameter to GlMapProvider; both callers already sit inside the provider !== 'leaflet' branch

**F134** CategoryManager.handleSave: removed the unreachable nameRequired guard (the key stays for MAdminCategoryManager). PackingTemplateManager.handleAddItem: removed !newItemName.trim(), the !expandedId part stays as a type narrow

**F135** Removed the leftover day-title rename machinery: startEditTitle, saveTitle, editingDayId/editTitle state, the focus effect, inputRef and the inline title <input>; onUpdateDayTitle stays declared in the props interface (no signature change) but is no longer destructured

**F136** Removed wouldBreakChronology and handleDropOnRow plus their hook returns and component destructures

**F137** Removed handleCalculateRoute, routeInfo and isCalculating state and the unused calculateRoute import; the day footer now renders only the plugin-minutes half; onRouteCalculated stays declared in the props interface

**F138** Collapsed the day Route toggle to two branches — the desktop 'select this day first' path could not be reached because the tools only render for the selected day

**F140** handleOptimize now takes a required `dayId: number` — the `= selectedDayId` default and the `if (!dayId) return` guard were unreachable

**F151** insertDay: removed the prevDays capture and set({ days: prevDays }) in the catch - the action never writes optimistically, so the rollback was unreachable.

**F154** buf = lines.pop() ?? '' -> lines.pop()! plus a brief comment; split('\n') always returns at least one element, so the ?? branch was unreachable.

**F160** Removed MobileStats (46 lines) from AtlasPage.tsx — no caller anywhere in the client, the phone path goes through MAtlas; the stale doc comment in useAtlas.ts was updated along with it

**F162** Removed the unused `const country = data?.countries.find(...)` in handleUnmarkCountry

**F165** Removed the guard line in the admin webhook test button that never fired (the button is already disabled without a URL); instead a comment explains why a masked value is deliberately tested without a url

**F172** Removed the local useIsMobile() ('(max-width: 720px)') from DashboardPage.tsx — below 767px DashboardPage routes to MDashboard. That drops the pass-card section, the !mobile branch of BoardingPassHero and the mobile branches of sideWidgets/sidebarVisible/AtlasStats

**F173** Removed the unreachable `if (!trip?.id) return` in removeCover; the remaining line reads the id directly in the edit branch

**F175** Removed availableUsers, getCountries and shareAvailableUsers from the local (non-exported) VacayApi interface and the api object; the sheets/panels talk to apiClient directly.

**F180** Identical to F113 — StatsLabel/buildStatsHtml removed; showStats stays declared as a prop (the prop signature is off limits) and is now commented as pure call-site compatibility instead of being silently ignored.

**F181** Identical to F114 — the stats leftovers are removed; showStats stays in ReservationOverlayOptions (exported contract) and now carries the explanatory comment.

**F182** Identical to F115 — escAttr and ZoomTracker are gone.

**F185** deleted handleDismiss/handleDismissById, the noticeIdRef that fed them, and the never-called onDismiss prop from ContentProps/NoticeContent/makeContentProps

**F187** removed the unreachable `if (!cta) handleDismissAll()` branch; runCta now takes NonNullable<cta> and the two handlers assert the invariant the render already guarantees

**F188** dropped the unused isRtlLanguage import, the language prop threading through BannerRenderer/AnimatedBannerItem/BannerItem, and the stale useCallback import

**F189** CTALink receives the cta as its own prop, which removed both unreachable `!notice.cta` guards

**F190** Identical to F160 — MobileStats is deleted, AtlasPage.tsx has no dead statements/branches left

**F193** removed the unreachable `if (!trip?.id) return` in handleRemoveCover

**F194** removed the unreachable `if (!user) return null` in the selected-members chips and `if (!value) return` in the member CustomSelect onChange

**F195** removed the unreachable `if (!inviteUrl) return` in copy() and `if (!selectedUserId) return` in handleAdd

**F202** CollabChatMessages.tsx: removed the unused `myReaction` along with the block body of the reactions map; ReactionBadge computes it itself.

**F203** CollabPolls.tsx: removed `if (!isClosed)` in onMouseEnter and the `!isClosed &&` guard in onClick — React dispatches neither onClick nor onMouseEnter on disabled form controls; a comment explains it.

**F205** Removed the four dead t(key) || 'literal' fallbacks (CollabPanel.tsx:58, CollabPolls.tsx:101 and 477, CollabNotes.tsx:274); in CollabPanel additionally the fallback field in ALL_TABS that became unused. All four keys exist in shared/src/i18n/en/collab.ts, so no new i18n keys were needed.

**F206** PackingShareControl.tsx: removed the dead ternary in onMouseLeave — the `if (!active)` guard already proves active === false.

**F207** PackingListPanelCategoryGroup.tsx: removed the unreachable catch blocks of handleCheckAll/handleUncheckAll (togglePackingItem rolls back and notifies on its own); a comment documents the responsibility.

**F208** PackingListPanelCategoryGroup.tsx: removed the duplicate `if (newItemName.trim())` in the add-item button's onClick; the button is disabled when the field is empty.

PR #1743, opened 2026-07-29
Add a dayTintProvider plugin hook

RenzoBeux:
## Why

Plugins can add rows to the day plan, but nothing can colour the day itself. A plugin that models a trip as a sequence of legs or phases can announce a transition as a row, but day 12 still looks exactly like day 40, so you can't tell where you are while scrolling a long itinerary.

`dayScheduleProvider` can't cover this. It renders a fixed row (a tinted icon plus text), and it's capped at 60 items per provider. One item per day would tint the first 60 days of a six-month trip and then silently stop, which fails exactly the long multi-destination trips this would be for. A separate hook is bounded by day count instead.

## What

`GET /api/day-tints/:tripId`, mirroring the day-schedule hook end to end (same gates, caps, fail-safe, per-provider budgets).

A day card has three separately tintable regions, so a plugin can mark a day boldly on the badge and leave a packed activity list plain:

| field | region |
|---|---|
| `badgeTone` / `badgeColor` | the day-number badge, and the mobile day chip |
| `headerTone` / `headerColor` | the header row (number, title, date, cost) |
| `activityTone` / `activityColor` | the expanded activity list |

`tone` and `color` are shorthands that fill every region you don't name. A region you never name isn't tinted and renders exactly as it does today.

A region takes either a tone from the shared palette or the plugin's own `#rrggbb` — four tones can't keep a twenty-stop trip's legs apart, which is most of the point of colour-coding one. The plugin picks the hue, the host picks the weight: a colour rides the same per-theme, per-region alphas as a tone (`#4F46E5` is itself a dark colour and vanishes on the dark sidebar at a light theme's alpha, and a large surface behind dense text needs a fainter tint than a small badge), and its OKLCH lightness is clamped into a per-theme band first, so no contribution can hand a user a day card they can't read. The band brackets the built-in tones rather than tightening around them, so an ordinary colour comes through untouched and a hex equal to a tone renders exactly like that tone.

A region ends up with exactly one paint instruction — what it names beats the shorthand, and a colour beats a tone at the same level — so the client never has a tie to break.

One contribution per day, resolved whole: first entry wins within a provider, first granted provider wins across plugins. A losing plugin can't fill in the winner's untinted regions, so two plugins can never each own part of one card.

## Notes

- This triggers re-consent. Adding a `hook:*` permission to an installed plugin means it won't activate until an admin re-approves.
- Only `#rrggbb` survives normalization, and that's a security boundary rather than input tidying. The value ends up inside a `color-mix()` in an inline `background`, which is a shorthand that takes a layer list, so a string closing the paren early could append a `url(...)` and beacon everyone viewing the trip. A region that names a colour the host can't use degrades to `default`, like a bogus tone already did.
- The clamp bounds a colour, it doesn't boost one. Hue and chroma are never touched, so a near-white grey still lands as a faint tint. That's the intended trade: the failure worth preventing is an unreadable card, not an understated one.
- A tinted header hovers to a deeper mix of its own colour rather than the neutral `--bg-tertiary`. Otherwise hovering flashes the day grey and reads as losing its colour. Selection still wins outright.
- The badge switches to `--text-secondary` when tinted. It's 11px bold, so 4.5:1 applies, and `--text-muted` on the untinted pill is already borderline at 4.34:1.
- With no tint-providing plugin installed, the sidebar renders identically to before.

## Tests

20 controller tests covering gates, caps, foreign dayIds, shorthand/override/named-only resolution, bogus tones, rejected colours (including a `url(...)` injection payload), cross-channel precedence, and both day-precedence rules — plus grant-mapping and SDK manifest cases. 6 client tests on the rendered `color-mix`, the lightness clamp and the untinted fallback. Server, plugin-sdk and client suites pass; tsc and lint clean on all three.

Since no test can prove the generated CSS actually resolves, I also checked the emitted strings in Chrome: all variants compute, and a near-black colour on the dark theme comes back at `srgb 0.145` against a `0.075` base instead of disappearing.


RenzoBeux:
Hi! As an example here are some screenshots of my plugin which I'm working on that would benefit from this provider.

Screenshot 2026-07-29 120400
Screenshot 2026-07-29 105641

image

RenzoBeux:
Tied to this plugin: https://github.com/liketrek/TREK-Plugins/pull/48

mauriceboe (changes requested):
Nicely done — this follows the existing provider-hook pattern closely, the Nest side is clean (injected `DatabaseService`, config via `app-config`, bridge for the non-Nest consumers), and 20 server unit tests including a `url()` injection payload is more than most hook PRs arrive with.

Two things before this can go in:

**1. SDK and runtime types disagree.** `plugin-sdk/src/index.ts:608` declares `color`, `badgeColor`, `headerColor` and `activityColor`, but `server/src/nest/plugins/runtime/plugin-sdk.ts:553-563` doesn't have them, and its comment says "Tone only, never a hex" — which contradicts the SDK. As it stands the main feature isn't type-safe for plugin authors. Please sync the two and fix the comment.

**2. Rebase needed.** Conflict is a single line in `plugins.module.ts:58-59`, purely mechanical.

One thing I want to flag as a deliberate decision rather than a defect: the three regions (`badge`, `header`, `activity`) turn the internal structure of the desktop day card into public plugin API. Once plugins depend on it, that layout is frozen. Mobile only honours `badgeTone` (`MTripShell.tsx:319`), so a plugin tinting headers looks broken on phones. I'm inclined to accept this, but I'd like it documented in the hook's doc comment so authors know what they're getting.

Also note each new `hook:*` permission forces re-consent on every installed plugin — worth a line in the release notes.

Could you get back to me within **3 days**? If there's no response by **6 August** I'll close it to keep the queue current; reopening once you have time is no problem.

RenzoBeux:
Rebased onto dev and pushed.

- The runtime copy now carries the four colour fields — the second commit updated the real SDK and missed the in-repo mirror, stale "tone only" comment included. Both are synced now.
- Documented the region caveat in the SDK doc comment (and the cookbook): the regions follow the desktop day card, and mobile honours only the badge, as the day chip — so the badge/shorthands are the one tint every user sees, and the finer regions are desktop refinement.
- Agreed on the release-notes line for the re-consent; happy to add it wherever the notes get drafted.

PR #1777, opened 2026-08-04
fix(admin): stop a stale base URL from hijacking the Anthropic endpoint

### LINKED ISSUE #1748: [BUG] Stale baseUrl hijacks the Anthropic endpoint
### Pre-flight checklist

### TREK version

3.4.1

### Describe the bug

After configuring the `local` provider with a base URL, switching the provider to `anthropic` and saving leaves the old base URL in the stored config. From then on, every Anthropic request goes to that host instead of `api.anthropic.com`. There is no indication of this anywhere in the UI or the logs.

The symptom is a series of errors that look like genuine Anthropic API responses but are not:

```
[llm-parse] AI parsing failed (provider=anthropic): Anthropic request failed (404):
{"type":"error","error":{"type":"not_found_error","message":"model 'claude-haiku-4-5-20251001' not found"},"request_id":"req_…"}
```

The same model ID succeeds when called directly against `api.anthropic.com` with the same API key:

```
$ wget -qO- https://api.anthropic.com/v1/models --header='x-api-key: sk-ant-…' --header='anthropic-version: 2023-06-01'
{"data":[{"type":"model","id":"claude-sonnet-5",…},{"id":"claude-haiku-4-5-20251001",…}], …}
```

With a PDF (native document path), the same misrouting produces an HTML error page from a local nginx, which the client surfaces verbatim as an "Anthropic" failure:

```
[DEBUG] Extracted (native PDF, 78455 bytes) sent to anthropic: Booking.com_ Conferma.pdf
[llm-parse] AI parsing failed (provider=anthropic): Anthropic request failed (400): <html>
<head><title>400 Request Header Or Cookie Too Large</title></head>
…
<hr><center>nginx</center>
```

The Anthropic API never returns HTML, which is what made the true cause identifiable.

### Root cause

In `client/src/components/Admin/AddonManager.tsx`, component `LlmParsingConfig`:

- The `baseUrl` state is initialised from the stored config and is not reset when `provider` changes.
- The Base URL input is only conditionally rendered (`{provider !== 'anthropic' && …}`), so the value becomes invisible but stays in state.
- `save()` unconditionally sends it:

```ts
await adminApi.updateAddon(addon.id, {
  config: { provider, model: model.trim(), baseUrl: baseUrl.trim(), apiKey, multimodal: cfg.multimodal === true }
})
```

Server side, in `server/src/nest/llm-parse/clients/anthropic.client.ts`, `AnthropicClient.extract()` honours it:

```ts
const base = (input.baseUrl ?? 'https://api.anthropic.com').replace(/\/+$/, '');
const url = `${base}/v1/messages`;
```

### Steps to reproduce

1. Admin -> Addons -> AI Parsing. Select provider `Local`, set any Base URL, save.
2. Switch provider to `Anthropic`, set a valid API key and a valid model ID, save.
3. Import any document. The request goes to the Base URL from step 1, not to `api.anthropic.com`.
4. Inspect the stored config to confirm: `SELECT config FROM addons WHERE id = 'llm_parsing';

### Expected behavior

The per-user equivalent already handles this correctly. In `client/src/components/Settings/LlmConnectionSection.tsx`:

```ts
const showBaseUrl = provider === 'local' || provider === 'openai'
…
llm_base_url: showBaseUrl ? baseUrl.trim() : '',
```

The admin panel is simply missing the same guard.

### Suggested fix

In `AddonManager.tsx`, mirror the per-user behaviour in `save()`:

```ts
baseUrl: provider === 'anthropic' ? '' : baseUrl.trim(),
```

`llm-config.resolver.ts` already maps an empty string to `undefined`, so the client falls back to the default endpoint with no further change.

Optionally, `AnthropicClient` could log the resolved base URL at debug level so a misrouted request is diagnosable from the logs alone.

### Deployment method

Docker Compose

### Host OS

Ubuntu 24.04 on proxmox VM 

### Accessing TREK from

Desktop browser

### Browser (if applicable)

Edge, chrome

```shell
[llm-parse] AI parsing failed (provider=anthropic): Anthropic request failed (404):
{"type":"error","error":{"type":"not_found_error","message":"model 'claude-haiku-4-5-20251001' not found"},"request_id":"req_…"}
```

### Screenshots

_No response_

### Additional context

All debugging executed with Claude code, I was investigating why I was having problems parsing PDFs with Anthropic and asked to inspect the code. Hopefully the solution proposed is acceptable, otherwise feel free to delete the issue

subdee:
## Description
`LlmParsingConfig.save()` in `client/src/components/Admin/AddonManager.tsx:402` sent
`baseUrl` unconditionally. The Base URL field is only rendered for non-Anthropic
providers (line 435), so switching a configured Local/OpenAI provider to Anthropic
hides the input but leaves the old URL in state — and it was still saved. The server's
`AnthropicClient.extract()` honours any `baseUrl`, so every Anthropic request was then
misrouted to that stale host instead of `api.anthropic.com`, surfacing as bogus
"Anthropic" 404s / nginx HTML errors.

Fix: gate the saved value on the provider — `baseUrl: provider === 'anthropic' ? '' :
baseUrl.trim()` — mirroring the field's own render condition and the already-correct
per-user path (`LlmConnectionSection.tsx`, `showBaseUrl`). This belongs in the admin UI
save payload, not the server: `llm-config.resolver.ts` already maps an empty string to
`undefined`, so `''` falls back to the default endpoint with no server-side change.

Closes #1748

## Type of Change
- [x] Bug fix

## Checklist

### DIFF
--- a/client/src/components/Admin/AddonManager.test.tsx
+++ b/client/src/components/Admin/AddonManager.test.tsx
@@ -560,6 +560,34 @@ describe('AddonManager', () => {
     });
   });
 
+  it('FE-ADMIN-ADDON-029: switching to Anthropic clears a stale base URL before saving', async () => {
+    const user = userEvent.setup();
+    const bodies: unknown[] = [];
+    server.use(
+      addonsRoute([llmAddon({ provider: 'local', model: '', baseUrl: 'http://ollama.lan:11434/v1', apiKey: '', multimodal: false })]),
+      modelsRoute([]),
+      http.put('/api/admin/addons/llm_parsing', async ({ request }) => {
+        bodies.push(await request.json());
+        return HttpResponse.json({ success: true });
+      }),
+    );
+    render(<><ToastContainer /><AddonManager /></>);
+
+    await screen.findByText('Installed on the server');
+
+    await user.click(screen.getByRole('button', { name: /Local · OpenAI-compatible/ }));
+    await user.click(screen.getByRole('button', { name: 'Anthropic' }));
+    await user.type(screen.getByPlaceholderText('claude-opus-4-8'), 'claude-haiku-4-5-20251001');
+    await user.type(screen.getByPlaceholderText('sk-…'), 'sk-ant-live');
+
+    await user.click(screen.getByRole('button', { name: 'Save' }));
+    await screen.findByText('Saved');
+    // The stale local base URL must not ride along to Anthropic — it would hijack the endpoint.
+    expect(bodies[0]).toEqual({
+      config: { provider: 'anthropic', model: 'claude-haiku-4-5-20251001', baseUrl: '', apiKey: 'sk-ant-live', multimodal: false },
+    });
+  });
+
   it('FE-ADMIN-ADDON-027: an error frame in the pull stream aborts the pull and is reported', async () => {
     const user = userEvent.setup();
     server.use(
--- a/client/src/components/Admin/AddonManager.tsx
+++ b/client/src/components/Admin/AddonManager.tsx
@@ -399,7 +399,7 @@ function LlmParsingConfig({ addon }: { addon: Addon }) {
     setSaving(true)
     try {
       // Send the masked sentinel unchanged so the server keeps the stored key.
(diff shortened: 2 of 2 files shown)


jubnl (approved):
Lgtm, thx!

PR #1822, opened 2026-08-07
fix(auth): redirect authenticated users away from /login and /register

### LINKED ISSUE #1810: [BUG] Manually going to the login page won't redirect you to default route when already logged in
### Pre-flight checklist

### TREK version

4.0.0

### Describe the bug

Pretty much the title; when you are already logged in and then manually visit the login url (or go back in browser history), you'll get served the login page instead of being redirected to the app default routing as is common.

### Steps to reproduce

1. Log in to trek
2. Change the url path to `/login`

### Expected behavior

Being redirected back to the dashboard overview

### Deployment method

Other

### Host OS

Demo site

### Accessing TREK from

Desktop browser

### Browser (if applicable)

_No response_

```shell

```

### Screenshots

_No response_

### Additional context

_No response_

subdee:
## Description
`/login` and `/register` (both render `LoginPage`) had no guard for already-authenticated
visitors — `client/src/App.tsx:220,223` rendered the page unconditionally, and the only
redirect logic (`ProtectedRoute`, `RootRedirect`, `useLogin`) ran the other direction or
only on a login action. So a logged-in user typing `/login` or using the back button saw
the login form instead of being redirected.

Adds a `GuestRoute` guard mirroring the existing `RootRedirect`/`ProtectedRoute` pattern and
wraps both auth routes with it; it redirects an already-authenticated visitor to `/dashboard`
(honoring `?redirect=` with the same open-redirect guard `useLogin` uses). The flag is
captured at mount via `useRef`, so a fresh login — which flips `isAuthenticated` true while
the takeoff animation still plays on `/login` before `useLogin` navigates — is left alone.
Routing is the right layer for this, alongside the other guards, rather than inside `useLogin`.

Closes #1810

## Type of Change
- [x] Bug fix

## Checklist

### DIFF
--- a/client/src/App.test.tsx
+++ b/client/src/App.test.tsx
@@ -291,6 +291,34 @@ describe('Public routes', () => {
   })
 })
 
+// ── PublicRoute — redirect already-authenticated visitors ─────────────────────
+
+describe('PublicRoute — already authenticated', () => {
+  it('FE-COMP-APP-012a: /login redirects to /dashboard when already authenticated', async () => {
+    seedAuth({ isAuthenticated: true, user: buildUser() })
+    renderApp('/login')
+    await waitFor(() => expect(screen.getByText('Dashboard')).toBeInTheDocument())
+    expect(screen.queryByText('Login')).not.toBeInTheDocument()
+  })
+
+  it('FE-COMP-APP-012b: /register redirects to /dashboard when already authenticated', async () => {
+    seedAuth({ isAuthenticated: true, user: buildUser() })
+    renderApp('/register')
+    await waitFor(() => expect(screen.getByText('Dashboard')).toBeInTheDocument())
+    expect(screen.queryByText('Login')).not.toBeInTheDocument()
+  })
+
+  it('FE-COMP-APP-012c: /login with a ?redirect= target is left to useLogin (OAuth consent handoff), not bounced', async () => {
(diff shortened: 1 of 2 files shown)


jubnl:
Verify that this has no impact with MCP Oauth 2.1, I remember that I had to remove the auto redirect on /login cause there was some issues with the mcp oauth on the consent page which contains a redirect quera param and idk if there's tests for that

subdee:
> Verify that this has no impact with MCP Oauth 2.1, I remember that I had to remove the auto redirect on /login cause there was some issues with the mcp oauth on the consent page which contains a redirect quera param and idk if there's tests for that

I excluded the login route that contains a redirect param, which is what OAuth does. In this case, OAuth won't break and redirecting if already logged in should work properly.

mauriceboe (approved):
Good catch, and the mount-time snapshot of the auth flag is the right way to keep a fresh login from bouncing itself mid-animation. I pushed two corrections on top: the bounce goes to START_DESTINATION_ROUTE instead of a hardcoded /dashboard, so it honours the start-page preference like every other post-login navigation; and the carve-out now checks for any query rather than ?redirect= alone, because this page also takes ?invite=, ?oidc_code= and ?oidc_error= and an invite link opened from a signed-in account would have been swallowed. Thanks.

PR #1867, opened 2026-08-09
Show place details while searching

mauriceboe:
Closes #1430.

The place search gives you a name and an address, and until now that was all you got before deciding. @sierikov's point in #1430 was that the little circle thumbnail "tells nothing for a place you never saw before", and that is fair.

On desktop the add-place dialog now carries a **Place details** column on the left. Picking a search result fills it in: pictures of the place and a description, each with its source. No extra dialog, no extra click — if you already know the place you ignore the column and save exactly as before, which was the condition for doing this at all.

Clicking a picture makes it the place's thumbnail, so it replaces the automatic one everywhere: list, map marker, itinerary, PDF and shared trips.

### Free sources are the normal case, not the fallback

Most instances have no Google key, so Wikipedia and OpenStreetMap carry this on their own and Google is what an admin adds on top:

- Pictures from **Wikimedia Commons**, found by coordinate. `geosearch` already returned five files and we kept one, so the whole strip costs no extra call.
- Descriptions from the OSM `description` tag or from the Wikipedia article the place is tagged with. The article is resolved from that tag, not guessed from the name — guessing lands on the wrong article for anything ambiguous, and a confident description of somewhere else is worse than none. No tag, no description.
- With a key, up to three **Google Places** photos and the editorial summary on top.

`getPlaceDetailsExpanded` already fetched the editorial summary but nothing ever asked for it. I gave it its own lean field mask instead of reusing that one, because its mask drags `reviews` along, which moves the call into the pricier tier.

### Attribution

The server has been returning an `attribution` field for photos for a long time and no client ever rendered it. That is survivable while one picture is fetched automatically; it is not survivable once you hand someone a strip of CC BY-SA images to choose from.

So Commons metadata now includes the licence name, licence URL and file description page, every thumbnail is credited under it, and the credit is stored with the cached file and shown under the avatar in the inspector afterwards.

### Cost and privacy

- Pictures are copied into the existing photo cache, never hotlinked. Google's photo URLs expire — migration 107 exists to repair exactly that — and a provider URL in the strip would send every viewer's address to Google or Wikimedia.
- Cache keys are `<placeId>~pN`. Unpicked candidates are unreferenced, so the nightly sweep removes them, and `isGooglePlaceId` rejects that shape so a candidate key can never reach Google (which bills the 400 it answers with).
- Results cached for a week in `place_details_cache` under a third `expanded` value, and rate limited to 60/min per user.
- **No migration.** The new admin switch reads fail-open on purpose, so no backfill is needed.

### Admin

`Place Enrichment` in Admin → Settings → API Keys, on by default. The Google half additionally follows the existing `Place Photos` and `Place Details` switches. Off means the column shows a short note and makes no outbound calls.

The three toggles next to it moved to `ToggleSwitch` — they were bare buttons with an inline transform, no `aria-pressed` and no label.

### Also in here

- `place:updated` now asks the planner to reload accommodations when the image changed, so a new thumbnail shows on the hotel card without a reload. Narrowed to actual image changes so renaming a place does not trigger a refetch.
- A GIF hero reaches the PDF. Uploads accept gif, the PDF extension check did not, so it silently fell back to the category icon.
- `NOTICE.md` gained Wikimedia/Wikipedia and Google Places entries.
- Two dead paths in the place contract (`server/src/routes/places.ts`, `services/placeService.ts`).

### Not in here

Mobile. The phone route has its own sheet and the column is desktop-only, as agreed in the discussion.

Draft while it soaks on dev1.


mauriceboe:
Bemerkung: Momentan ist die Beschreibung der Orte fehlerhaft/wird nicht angezeigt

PR #1874, opened 2026-08-10
fix(feeds): use per-endpoint zones for transport times in the ICS feed

### LINKED ISSUE #1453: [BUG] Subscribed Calendar Events Not Respecting Times Zones

subdee:
## Description
`exportICS` in `server/src/nest/trips/trips.service.ts` resolved a reservation's
calendar times from `reservation_time` first, and only fell back to per-endpoint
zones when `reservation_time` was absent. That branch derives its zone from the
linked place — a flight has none — so `resolveTimeZone` returns null and both
DTSTART/DTEND are emitted as RFC-5545 floating times, which subscribers' calendars
render in their own local zone rather than TREK's.

The assumption that transports never carry a top-level `reservation_time` is wrong:
TransportModal stamps departure/arrival onto `reservation_time`/`reservation_end_time`
alongside the endpoints, so every flight took the floating-time branch.

Fix: consult a timed departure endpoint first, so DTSTART follows the departure
endpoint's IANA zone and DTEND the arrival endpoint's, each with its own TZID.
`reservation_time` remains the fallback for reservations with no timed endpoints
(hotels/restaurants, which correctly resolve a single zone from their place). A
timed reservation backed only by date-only endpoints still keeps its timed value.

Closes #1453

## Type of Change
- [x] Bug fix

## Checklist


mauriceboe (dismissed):

mauriceboe (approved):
The reordering is right: a flight's only zone on the reservation_time branch is the linked place, which a flight does not have, so both ends floated. I pushed one follow-up on top, keeping the DTEND from reservation_end_time when the arrival side has no clock of its own, so a one-sided rental does not lose its multi-day return. Thanks.

PR #1875, opened 2026-08-10
feat(POI): Adds plugin Hook for POI categories 

Arnvig:
## Description
I want to expand the POI category, so it is possible add and write plugin that adds a new category and POIs to the map.

It adds a new 'poiCategoryProvider' plugin hook, the server endpoint for plugin POI categories, SDK/server permission wiring, tests, wiki updates, and a demo plugin. I also verified it locally in Docker.

Addresses discussion: https://github.com/liketrek/TREK/discussions/1781

## Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [x] Documentation update

## Checklist


client/src/components/Admin/AdminPluginsPanel.tsx
```diff
@@ -233,6 +233,10 @@ function deriveCaps(perms: string[], caps: { widget?: { slot?: string }; tripPag
   if (perms.includes('hook:place-detail-provider')) out.push({ icon: MapPin, label: t('admin.plugins.cap.placeDetails') })
   if (perms.includes('hook:trip-warning-provider')) out.push({ icon: AlertTriangle, label: t('admin.plugins.cap.warnings') })
   if (perms.includes('hook:map-layer-provider')) out.push({ icon: Route, label: t('admin.plugins.cap.mapLayers') })
+  if (perms.includes('hook:poi-category-provider')) {
+    const poiLabel = t('admin.plugins.cap.poiCategories' as never)
+    out.push({ icon: MapPin, label: poiLabel === 'admin.plugins.cap.poiCategories' ? 'Provides map POIs' : poiLabel })
```
Arnvig, line 238: I'm unsure if this code is nessesary

mauriceboe:
comment dc: It might be to much AI-slam generated, and I'm normally a C# developer 😅 
So please be skeptical of the code.
 
The PR is here
https://github.com/liketrek/TREK/pull/1875

But I have teste it in a local docker, with one and two demo plugin running.

mauriceboe (changes requested):
Thanks for this, and thanks for being upfront about how it was written. The server half is genuinely good: the controller sits in the right module, injects `PluginHooks` instead of reaching for anything global, normalises everything the plugin sends back (string coercion, length caps, coordinate ranges, URL scheme allowlist, result cap) and comes with a real unit test. That is the shape I want contributions in this area to have.

The client half and the register bookkeeping are where I have to send it back.

### Three things that will fail CI as it stands

1. **`client/src/components/Map/usePoiExplore.ts`** (the `fetchCat` fan-out around line 135): `res` is a union of the plugin response and the maps response, and the code reads `res.providers` on one branch and `res.pois` on the other. TypeScript cannot narrow that from the independent `pluginMeta` condition, so `npm run typecheck` fails. Split it into two branches that each own their call and their result instead of sharing one `res`.

2. **`server/tests/unit/plugins/plugin-hooks.test.ts`**: `CONTRACTS` is compared for set equality against the declared hooks, so the two new `@PluginHook` declarations have to be added there or PLUGHOOK-002 goes red. That table is deliberately hand maintained: it is the place where adding a hook stops being a local change.

3. **i18n**: `admin.plugins.cap.poiCategories` only exists in `shared/src/i18n/en/admin.ts`. CI runs `node shared/scripts/i18n-parity.mjs --strict`, so a key that lives only in `en` blocks the PR. All 23 languages have to carry it.

### A bug worth fixing before the rest

The category lookup and the search run inside the same `Promise.all`. A plugin that declares the hook but has a broken or missing `search()` therefore takes the whole batch down with it, and disappears from the pill entirely, even though `providersOf` only ever promised that it reported the hook name. Settle those independently so one bad plugin degrades to "no results" instead of "no category".

While you are in there: the two hardcoded English fallbacks (`'Provides map POIs'`) can go. `t()` already falls back to `en`, so the defensive branches around them are dead.

### The part I would think about again

Right now every active plugin category triggers its own full fan-out across all providers, and the category list is fetched on mount whether or not the POI pill is even enabled. For a map that refetches on every viewport change that adds up fast. I would rather have the category list on its own cheap route (or a `?categoriesOnly=1`), have it depend on `poiPillEnabled`, and bundle one search request per viewport change instead of one per active category.

### Before the next round

Two process things. The title is still the template (`feat(component): short description of new feature`). And #1781 is an open feature request that nobody has answered yet, so there is no agreement that this particular cut is what we want. `mapMarkerProvider` already lets a plugin put points on the trip map, and I would like to decide deliberately whether we want a second way to do that before it ships, rather than after. Let's settle that on the discussion first, then the code.

Arnvig:
Thank you for the through review, I will take notes from the review.

Of course I will be upfront about how I code it :)

I will wait, unto it have been discussed - before I go further with second rund.

I may have taken on too big a task, when i'm not comfortable in TS yet...
So I will look into issues and approve things and se if there some other things I can help with :)

PR #1890, opened 2026-08-11
feat(dayplan): per-segment travel mode for boundary legs (#1281 follow-up)

geoida:
Follow-up to the per-segment travel mode from #1651. After living with that feature across a multi-city trip, the mode picker only attaches to a leg whose **origin is a place** — so four kinds of legs that are still drawn on the day plan stay read-only:

1. the first ground leg after a transport booking (ferry/train/bus arrival → next stop),
2. the same after a transit booking,
3. the evening accommodation bookend (last stop → the day's hotel),
4. the morning accommodation bookend (the day's hotel → first stop).

The mode of those legs couldn't be changed, and (1), (2) and (4) always drew in the day default because there was nowhere to store an override — `leg_transport_mode` on `day_assignments` means "the leg *leaving* this stop", and their origin isn't a place.

## Approach

A leg's mode is owned by whichever end is a real stop:

- origin is a place → its existing outgoing `leg_transport_mode` (unchanged),
- origin isn't a place but the destination is → a new **`incoming_leg_transport_mode`** on the destination stop.

That one nullable column covers cases 1, 2 and 4 (destination-owns-incoming); case 3 already had a place origin and just needed the connector wired up. Everything stays on `day_assignments`, so there's a single storage location and read path, and resolution is one shared helper (`resolveLegMode(origin, dest, dayDefault)`) used by both the sidebar and the map route-calc; the mobile timeline consumes the segments they produce.

## What's in it

- migration: `incoming_leg_transport_mode TEXT` (nullable, guarded) — existing itineraries are unchanged.
- read path: projected in both `getAssignmentWithPlace` and `rowShape.ts::formatAssignmentWithPlace` (the list projection).
- write path: `PUT /assignments/:id/transport` gains an optional `direction: 'outgoing' | 'incoming'` (defaults to `'outgoing'`, so existing callers are untouched); a `setIncomingLegTransportMode` service method mirrors the outgoing one.
- resolution: `resolveLegMode` in the day sidebar and the map route calc — boundary legs now draw in their resolved mode.
- UI: the mode menu is exposed on the booking connector (writes the next stop's incoming), the morning hotel connector (first stop's incoming) and the evening hotel connector (last stop's outgoing). The menus always target the leg's real located endpoint, so a coordinate-less edge stop or a booking edge gets no menu rather than a silent mis-write.
- tests: migration, list-endpoint parity, the `direction` API (default-outgoing back-compat, incoming write, invalid-direction 400), the resolver's semantics, and outgoing/incoming independence on one stop.

## Known limitation (intentional)

A leg with **no place endpoint on either side** still can't carry a mode — there's nowhere to store it:

- booking → booking (e.g. a ferry arrival straight into a train) isn't drawn today anyway (the `curHasPlace` gate), so no change there;
- hotel → booking (a morning hotel straight to a departure, no stop between) **is** drawn but stays in the day default, uneditable — its endpoint isn't a place, so the connector shows no mode menu. Uncommon but real; left as an inert leg rather than growing the schema for it.

A follow-up PR will add a "public transport" option to the same per-segment menu that pre-fills the transit-journey search from the leg (origin/destination/departure); it builds on this one's connector-menu plumbing, so it'll come after this lands.

### DIFF
--- a/client/src/api/client.ts
+++ b/client/src/api/client.ts
@@ -455,7 +455,9 @@ export const assignmentsApi = {
   setParticipants: (tripId: number | string, id: number, userIds: number[]) => apiClient.put(`/trips/${tripId}/assignments/${id}/participants`, { user_ids: userIds } satisfies AssignmentParticipantsRequest).then(r => r.data),
   updateTime: (tripId: number | string, id: number, times: AssignmentTimeRequest) => apiClient.put(`/trips/${tripId}/assignments/${id}/time`, times).then(r => r.data),
   // Per-segment travel mode (#1281): mode of the leg leaving this stop (null = inherit day default).
-  updateTransport: (tripId: number | string, id: number, mode: string | null, direction: 'outgoing' | 'incoming' = 'outgoing') => apiClient.put(`/trips/${tripId}/assignments/${id}/transport`, { transport_mode: mode, direction } satisfies AssignmentTransportRequest).then(r => r.data),
+  // direction defaults to 'outgoing' server-side, so only send it for the incoming (boundary-leg) case
+  // and keep the outgoing payload byte-for-byte identical to the pre-#1281 shape.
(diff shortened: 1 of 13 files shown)


mauriceboe:
thanks <3

PR #1893, opened 2026-08-11
feat(atlas): show wishlist countries as hatched, stable colors, remove action

NtsCiccio:
## Description
- Countries that are on the wishlist (bucket list) but not yet visited are now
  rendered on the Atlas map with a diagonal-stripe hatch pattern, instead of
  blending into the flat gray used for unvisited countries.
- Country colors are now deterministic, derived from a hash of the ISO A3 code
  (`countryColor` in atlasModel.ts) instead of the visit-order index. Previously,
  marking a new country as visited (or otherwise changing the order) could
  reshuffle the colors of every other country already shown on the map —
  a confusing visual bug.
- Added a "Remove from wishlist" action to the country selection popup, so a
  country can be taken off the bucket list directly from there without
  switching to the Bucket List tab.
- Updated i18n strings (`atlas.removeFromBucket`, `atlas.removeFromBucketHint`).
- Added/updated unit tests (atlasModel.test.ts) and integration tests
  (AtlasPage.test.tsx) covering the hatch pattern, color stability, and the new
  remove action.

## Type of Change
- [x] Bug fix (unstable country colors reshuffling)
- [x] New feature (wishlist hatch pattern + remove action)
- [ ] Breaking change
- [ ] Documentation update

### DIFF
--- a/client/src/mobile/screens/atlas/MAtlasCountryPopup.test.tsx
+++ b/client/src/mobile/screens/atlas/MAtlasCountryPopup.test.tsx
@@ -48,7 +48,7 @@ describe('MAtlasCountryPopup', () => {
     const handleDeleteBucketItem = vi.fn().mockResolvedValue(undefined)
     const setConfirmAction = vi.fn()
     const atlas = makeAtlas({
-      // Two places in France plus one elsewhere — only the French ones go.
+      // Two places in France plus one elsewhere, only the French ones go.
       bucketList: [bucketItem(1, 'Paris', 'FR'), bucketItem(2, 'Nice', 'FR'), bucketItem(3, 'Tokyo', 'JP')],
       handleDeleteBucketItem,
       setConfirmAction,
--- a/client/src/mobile/screens/atlas/MAtlasCountryPopup.test.tsx
+++ b/client/src/mobile/screens/atlas/MAtlasCountryPopup.test.tsx
@@ -1,0 +1,76 @@
+// FE-COMP-MATLASPOPUP-001 to FE-COMP-MATLASPOPUP-004
+import React from 'react'
+import { describe, it, expect, vi } from 'vitest'
+import { render, screen, fireEvent, waitFor } from '../../../../tests/helpers/render'
+import MAtlasCountryPopup from './MAtlasCountryPopup'
+import type { AtlasController } from './atlasController'
+
+const bucketItem = (id: number, name: string, countryCode: string | null) => ({
+  id, name, country_code: countryCode, lat: null, lng: null, notes: null, target_date: null,
+})
+
+// Only the slice of the atlas hook this sheet reads. `t` echoes the key back so
+// the assertions name the contract, not a locale string.
+function makeAtlas(overrides: Record<string, unknown> = {}): AtlasController {
+  return {
+    t: (key: string) => key,
+    confirmAction: { code: 'FR', name: 'France', type: 'choose' },
+    setConfirmAction: vi.fn(),
+    executeConfirmAction: vi.fn(),
+    setData: vi.fn(),
+    setVisitedRegions: vi.fn(),
+    setBucketList: vi.fn(),
+    visitedRegions: {},
+    bucketList: [],
+    handleDeleteBucketItem: vi.fn().mockResolvedValue(undefined),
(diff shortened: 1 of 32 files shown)


NtsCiccio:
@mauriceboe 
Thanks for reviewing the PR. I just wanted to let you know that I love Trek and am really happy to have made my small contribution. If there are other things you need help with, I'm available—Trek was the first app I ever self-hosted.

mauriceboe:
Hi @NtsCiccio! Thanks for your idea! I had to make a few adjustments because you'd left out a few things and all the tests were failing. But it's in there, and I hope you're looking forward to version 4.0.0! :)

NtsCiccio:
> Hi @NtsCiccio! Thanks for your idea! I had to make a few adjustments because you'd left out a few things and all the tests were failing. But it's in there, and I hope you're looking forward to version 4.0.0! :)

If you don't mind (I really appreciate feedback to help me improve), could you tell me where I went wrong and what I could have done better?
That way, I hope to do better next time.
Anyway, yeah, I can't wait for 4.0.0. As I said, I love this project, and I hope I can be helpful with some issues or bug fixes—even “first-time issues” or whatever they're called now.

mauriceboe:
Happy to, and let me start with the part that matters most: the substance of this PR was right.

Spotting that the colours were assigned by index into the visited list, and that this reshuffles the whole map whenever one more country is marked, is a proper bug find. Hashing the ISO code instead is exactly the fix I would have picked. And you did the thing most people skip: you pulled the region layer onto the same function instead of leaving a second copy of the palette behind. The `CanvasPattern` for the hatch with a `null` fallback for jsdom shows you were thinking about the tests while writing it, which is not a beginner move.

So, the three things I patched up after merging, in the order they would have cost you time:

**1. Translations, all 23 of them.** You added `atlas.removeFromBucket` and its hint to `en` only. Every locale in `shared/src/i18n/` has to carry the same key set or the *i18n Key Parity* job fails, which is what turned the PR red. You can catch it in a second locally:

```
npm run i18n:parity:strict --workspace=shared
```

Don't worry about the translations being perfect. Matching the wording each locale already uses for the surrounding feature is plenty, and native speakers correct the rest later.

**2. An existing test held the old behaviour.** `useAtlas.test.tsx` had `FE-HOOK-ATLAS-035`, which asserted that France gets `#6366f1`, the first palette entry, because it was the first visited country. Your change makes that false by design, so the test failed for the right reason. Whenever you change something central, it pays to grep for who else depends on it before pushing:

```
grep -rn "6366f1" client/src --include="*.test.*"
```

The fix was not to write the new hash value into the test, that would just freeze a different implementation detail. I rewrote it to assert the property the test actually cares about: France keeps the same colour before and after toggling the planned layer, and a planned country never takes it. Tests that describe behaviour instead of values survive the next refactor.

**3. Mobile.** The hatch and the stable colours showed up on phones for free, because `MAtlas` uses the same `useAtlas` hook you changed. The new *Remove from wishlist* action did not, because the phone has its own popup in `client/src/mobile/screens/atlas/MAtlasCountryPopup.tsx`. Rule of thumb: change a hook and mobile follows, add a button and it does not. Worth a look under `client/src/mobile/` whenever you add an action.

**A short list for next time**, all from the repo root:

```
npm run i18n:parity:strict --workspace=shared   # locale key parity
npm run test --workspace=client                 # or npx vitest run <file> while iterating
cd client && npm run theme:lint                 # flags hardcoded colours (not in CI, run it yourself)
```

None of this is a reason to hesitate. A PR that lands with a real bug fix and needs a translation sweep is a good trade for everyone, and this one is in 4.0.0.

As for what to pick up next: issues carrying **approved on discord** are cleared to work on, and any open bug report with a clear reproduction is fair game. If you want something bigger, pitch it in `#github-pr` on Discord first, that is in the contributing guide and it saves you from building something that then gets turned down. Genuinely glad to have you around.

NtsCiccio:
Thank you so much for the feedback—this is my first real contribution to an open-source project, and I’m thrilled.
I’ll keep your advice in mind. Thanks again. I’ll check out Discord (I’ve also started a thread about this issue there), and I’ll see if there’s anything else I can help with.

PR #1910, opened 2026-08-12
feat(dayplan): public transport option on per-segment connectors (#1281 follow-up)

geoida:
Follow-up to the boundary-leg travel modes in #1890, which promised this. That PR let every connector on the day plan carry a per-segment mode; this one adds the option that was missing from the menu: **"Public transit"**, so a leg can be planned as a real public-transport journey instead of an OSRM road route.

## What it does

The per-segment mode menu (the one that opens on a connector) gains a "Public transit" entry alongside the route profiles. Choosing it opens the existing transit search (#1065) pre-filled with the leg's own endpoints:

- **from** = the leg's origin,
- **to** = the leg's destination,
- **departure time** = the origin stop's time.

You compare itineraries and add one exactly as the day-header transit button already does; the saved journey slots into the timeline like any other. The entry appears on every editable connector: stop-to-stop legs and the boundary hotel/booking bookends from #1890 alike.

## Approach

A connector only renders once both of its endpoints are located, so "search public transit for this leg" is inherently limited to legs that can actually be searched: no separate gating, no menu on a coordinate-less edge. The connector carries only the leg's coordinates, so each end is resolved back to a named, timed point from the trip's own data (places, accommodations and booking endpoints) before seeding the search. That keeps MOTIS driven by the coordinates while the itinerary title and departure time read naturally.

Nothing new is stored and no contract changes: this reuses the transit search's existing `transitPrefill` path (from/to) and extends it with an optional departure time. `TransitSearchPanel` gains an `initialTime` prop; `TransportModal` and the planner state thread it through.

## What's in it

- `DayPlanSidebar`: a shared resolver from leg coordinates to `{ name, lat, lng, time }`, a "Public transit" menu entry wired onto both the outgoing and incoming connector menus, and an `onPlanTransitLeg` callback the page turns into a pre-filled transit search.
- `TransitSearchPanel`: `initialTime` seeds the departure time (defaults to the existing 09:00).
- `TransportModal` + planner state: carry the optional `time` on the transit prefill.
- tests: the stop-to-stop connector offers the option pre-filled with the leg's from/to/time; a hotel bookend connector offers it from the hotel; the entry is absent without the handler; `initialTime` seeds the search.

Scoped to the client; no schema, API or i18n changes (the menu label reuses the existing `transit.title`). Client typecheck clean, lint clean, the affected suites pass.


mauriceboe (approved):
Reviewed against dev. This is the follow-up #1890 announced in its own description, and it holds together end to end: the saved journey keeps the picked coordinates as its endpoints, so planDay breaks the route exactly there and the road connector disappears instead of sitting next to the transit row. The coordinate lookup is safe because calculateRouteWithLegs returns the input waypoints rather than the OSRM-snapped ones, and the test mock was corrected to match reality rather than the assertion loosened. Nice catch on your own second commit with the cross-day time leak.

No blockers. One thing for later, not for this PR: the phone has its own per-leg menu in MPlanTimeline and does not get the entry, so mobile stays without it for now.

PR #1913, opened 2026-08-12
fix(atlas): stop the panel clipping the top of the bucket search results

### LINKED ISSUE #1899: [BUG] First bucket item search result is cut off within Atlas
### Pre-flight checklist

### TREK version

4.0.0

### Describe the bug

When manually searching for a bucket list entry, the first result will be cut off when multiple results are returned.

### Steps to reproduce

1. Go to atlas, switch to bucket list items
2. Click the 'Add place' button
3. Search for `Amsterdam`

### Expected behavior

All results being visible and easily selectable

### Deployment method

Other

### Host OS

Demo env

### Accessing TREK from

Desktop browser

### Browser (if applicable)

Zen (firefox)

```shell
```

### Screenshots

Image

### Additional context

_No response_

subdee:
## Description
The bucket-list place search rendered its results with `position: absolute; bottom: 100%`
inside the desktop Atlas panel (`client/src/pages/AtlasPage.tsx:585`). That panel is
`overflow-hidden` (`AtlasPage.tsx:128`) for its rounded corners and glare layers, and it
leaves only ~125px of room above the input — so the top ~35px of the 160px list fell
outside the clip box. Because that is the container's own top rather than a scroll offset,
the first result could not be seen or scrolled to, only guessed at.

The list is now portalled to `document.body` and positioned `fixed` from the input row's
bounding rect, flipping up or down depending on available room and capping its height to
that room so it scrolls rather than running off the viewport. This is the pattern already
used by `CustomSelect` (`client/src/components/shared/CustomSelect.tsx:121`) — whose
month/year dropdowns sit in this very form and already escape the panel this way — so the
fix stays in the presentation layer and adds no new machinery.

Two existing tests asserted the old `position: absolute` layout; both are updated. One of
them (`FE-PAGE-ATLAS-038`) would otherwise have degraded to a vacuous pass, since its
locator is guarded by `.catch(() => null)`.

Closes #1899

## Type of Change
- [x] Bug fix

## Checklist
- [χ] I have tested my changes locally

### DIFF
--- a/client/src/pages/AtlasPage.test.tsx
+++ b/client/src/pages/AtlasPage.test.tsx
@@ -922,6 +922,41 @@ describe('AtlasPage', () => {
     });
   });
 
+  describe('FE-PAGE-ATLAS-051: bucket search results escape the panel that clips them', () => {
+    it('renders the result list outside the overflow-hidden desktop panel (#1899)', async () => {
+      server.use(
+        http.post('/api/maps/search', () => HttpResponse.json({
+          places: [
+            { name: 'Amsterdam', address: 'Amsterdam, North Holland, Netherlands', lat: 52.37, lng: 4.89 },
+            { name: 'New Amsterdam Island', address: 'French Southern and Antarctic Lands, France', lat: -37.8, lng: 77.5 },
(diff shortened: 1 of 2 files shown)


jubnl (approved):
Lgtm

PR #1919, opened 2026-08-13
feat(feed): car rental pickup and drop-off events in the ICS feed

michael-bohr:
## Description

Car rentals now emit **Pickup** and **Drop-off** as their own timed events in the ICS feed, so a subscribed calendar shows when to collect the car and when it is due back instead of one opaque block.

This is the last unported piece of #1721. Everything else in that PR has since landed independently on `dev`: hotels got the all-day stay range plus timed check-in/check-out (including `check_in_end` as the check-in window's DTEND) in #1915, and the flight floating-time regression was fixed with the departure-endpoint-first logic citing #1453. Rather than rebase a diff that is modify/delete against three files that no longer exist, this is a fresh, much smaller change against the code's current home in `calendar.service.ts`.

Deliberately additive: the rental's existing single-block event is untouched, exactly as the check-in/check-out markers are additive to the all-day stay range.

**How the two sides are derived**, in priority order:

1. Endpoint `role` (`from` = pickup, `to` = drop-off), reusing the endpoints map already loaded in this method.
2. If no endpoint carries either role, positional first/last by `sequence` — and only when there is more than one endpoint, so a lone endpoint is never treated as both sides. An import that drops one endpoint to failed geocoding must not let the surviving return endpoint masquerade as the pickup.
3. Otherwise `reservation_time` / `reservation_end_time`, since imports frequently carry only that window. `reservation_end_time` is often a bare clock, so the drop-off date falls back to the reservation's own date.

A side emits nothing unless a usable date **and** clock resolve. Zone resolution per side prefers the endpoint's stored `timezone`, then its coordinates, then the reservation's linked place, matching the surrounding code.

Follows the hotel `marker` closure's pattern closely — same `esc`/`uid`/`dtLine`/`isDate`/`isTime`/`usedZones` usage, same `Pickup:`/`Drop-off:` summary convention mirroring `Check-in:`/`Check-out:`, same per-kind UID shape.

Addresses the remaining scope of #1721, which you reviewed with "This is the most valuable PR in my current queue, and I want it in."

Two notes carried over from that review:

- **Metadata fallback left out on purpose.** Your note asked for `acc.check_in` before `meta.check_in_time`, but the merged `marker` reads the stay values only and never touches metadata, so there is no precedence left to invert. Happy to add metadata strictly *below* the stay values in a follow-up if you want bookings whose clock only lives there to produce events — just say the word.
- **`TRANSP:TRANSPARENT`** on the trip banner still looks open, and you said that one was yours.

## Type of Change

- [x] New feature

## Checklist

## Testing

`CAL-032` through `CAL-036` added to `server/tests/unit/nest/calendar.service.test.ts`, covering: both sides from `from`/`to` endpoints at the right local times and zones; role-less endpoints falling back to first/last by sequence; a rental carrying only `reservation_time`/`reservation_end_time`; a single endpoint not producing a bogus second event; and a rental with no usable clock emitting nothing.

Full calendar suite passes 54/54 (49 pre-existing untouched, including the hotel `CAL-025`/`026`/`027`), `tsc --noEmit` clean.

Happy to add a line to `wiki/Calendar-Feeds.md` if you'd like this documented there.

### DIFF
--- a/server/src/nest/calendar/calendar.service.ts
+++ b/server/src/nest/calendar/calendar.service.ts
@@ -182,10 +182,15 @@ export class CalendarService {
         // Pad to 15 chars (YYYYMMDDTHHMMSS) — add missing seconds
         return raw.length === 13 ? raw + '00' : raw;
       }
-      // Time-only: combine with reference date
+      // Time-only: combine with reference date. Pad the same way the branch above
+      // does rather than unconditionally, because the value can already carry
+      // seconds ("10:00:00" is what the booking import stores). Appending
+      // regardless made it 17 characters, which fails dtLine's shape check, drops
+      // the TZID and leaves a floating time in the feed (#1453 again).
       if (refDate && d.match(/^\d{2}:\d{2}/)) {
         const datePart = refDate.split('T')[0];
-        return `${datePart}T${d.replace(/:/g, '')}00`.replace(/-/g, '');
+        const raw = `${datePart}T${d.replace(/:/g, '')}`.replace(/-/g, '');
+        return raw.length === 13 ? raw + '00' : raw;
       }
       return d.replace(/[-:]/g, '');
     };
--- a/server/src/nest/calendar/calendar.service.ts
+++ b/server/src/nest/calendar/calendar.service.ts
@@ -483,6 +483,77 @@ export class CalendarService {
       }
     }
 
+    // Pickup and drop-off as their own timed events — the car-rental analogue
+    // of check-in/check-out above (last unported piece of #1721). Additive,
+    // same as the check-in/check-out markers are additive to the all-day stay
+    // range: the rental's existing single-block event (from the endpoint or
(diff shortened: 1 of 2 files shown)


mauriceboe (approved):
Good work on this one: no extra query, it reuses the endpoints map already in hand, the UID namespaces cannot collide, and the guard against a lone endpoint being reused as both sides is exactly the right instinct for a half-geocoded import.

One thing needed fixing and I pushed it rather than sending you back a round: fmtDateTime's time-only branch appended '00' for seconds no matter what, so a clock that already carried them came out 17 characters, dtLine refused the TZID and the event went out floating. That is the form the booking import stores, so it would have hit real rentals. Fixed at the cause so every caller benefits, with CAL-042 pinning it.

Thanks also for rebasing onto dev and renumbering the test ids yourself in the meantime, that collision was caused by my own PR landing first.

PR #1922, opened 2026-08-13
feat: enable environment proxy support in container

Cynosure159:
﻿  ## Description

  Enable environment-based outbound proxy support for TREK container deployments.

  Node.js `fetch` does not use `HTTP_PROXY`, `HTTPS_PROXY`, or `NO_PROXY` by default. This PR enables Node.js environment proxy
  support in the container image by setting `NODE_USE_ENV_PROXY=1`.

  It also:

  - Adds optional, commented proxy variables to `docker-compose.yml`
  - Documents `HTTP_PROXY`, `HTTPS_PROXY`, and `NO_PROXY`
  - Keeps outbound proxying disabled unless proxy variables are explicitly configured

  Requests handled by TREK's SSRF protection use a dedicated dispatcher and are not affected by the environment proxy.

  ## Type of Change

  - [x] Bug fix
  - [ ] New feature
  - [ ] Breaking change
  - [x] Documentation update

  ## Checklist

  - [x] My branch is [up to date with `dev`](https://github.com/liketrek/TREK/wiki/Development-environment#3-keep-your-fork-up-to-
  date)

  ## Testing

  - Successfully built the container image for `linux/amd64` and `linux/arm64`
  - Verified the multi-architecture image manifest
  - Verified that proxy configuration remains disabled by default
  - Verified Compose and Markdown changes with `git diff --check`

  Automated tests were not added because this change only enables Node.js's built-in environment proxy support at the container
  runtime level.

Addresses discussion #1754 ("Add Proxy Settings"), which reports exactly this: `HTTP_PROXY` / `HTTPS_PROXY` set on the container have no effect, because Node ignores them unless it is started with `NODE_USE_ENV_PROXY=1`.

_Added by a maintainer while triaging: the reference was missing from the original description._

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -114,6 +114,7 @@ RUN mkdir -p /app/data/logs /app/uploads/files /app/uploads/covers /app/uploads/
     chown -h node:node /app/server/uploads /app/server/data
 
 ENV NODE_ENV=production
+ENV NODE_USE_ENV_PROXY=1
 ENV PORT=3000
 ARG APP_VERSION=dev
 ENV APP_VERSION=${APP_VERSION}
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -31,6 +31,9 @@ services:
 #      - COOKIE_SECURE=false # Escape hatch: force session cookies over plain HTTP even in production. Not recommended.
 #      - TRUST_PROXY=1 # Trusted proxy count for X-Forwarded-For / X-Forwarded-Proto. Required for FORCE_HTTPS to work.
 #      - ALLOW_INTERNAL_NETWORK=false # Set to true if Immich or other services are hosted on your local network (RFC-1918 IPs). Loopback and link-local addresses remain blocked regardless.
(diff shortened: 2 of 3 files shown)


mauriceboe (approved):
Verified rather than taken on trust: with NODE_USE_ENV_PROXY=1 and a proxy set, a plain fetch goes through it; with the flag alone and no proxy variables, it is a genuine no-op. A request through the SSRF guard's pinned dispatcher still goes out directly, so the DNS pinning stays intact, and the plugin sandbox never sees the variables because its env is a whitelist. Good, minimal choice: the runtime's own flag rather than a hand-rolled dispatcher.

I pushed two things onto the branch while reviewing, both documentation. The page also serves Helm, Unraid and source installs, where nothing sets the flag, so it now says the switch lives in the image. And NO_PROXY got a recommendation: without it the proxy sees every request TREK makes to itself, including the container health check. Also referenced discussion #1754 in the description, which is the report this fixes.

Thanks for this one.

PR #1929, opened 2026-08-13
feat(mcp): MCP tools for per-leg and day-default transport modes

geoida:
## Description

Follow-up to the per-segment travel modes (#1281) and the boundary-leg modes in #1890. Those let each leg of a day carry its own transport mode, and the whole day carry a default; but the only way to set those from an MCP client was a raw DB write. This PR closes that gap with two MCP tools that wrap the exact service methods the REST routes already use.

- **`set_leg_transport_mode`** – sets a leg's travel mode for a place assignment. `direction` `"outgoing"` (default) targets the leg leaving the stop; `"incoming"` targets the arriving leg for a non-place origin (a booking arrival or a morning hotel departure), mirroring the `PUT :tripId/assignments/:id/transport` route's `direction` branch.
- **`set_day_default_transport_mode`** – sets (or clears, with `null`) a day's whole-day default mode, mirroring `PUT :tripId/days/:id/transport`.

Both are pure adapters: no new business logic, no schema or API change. They call `AssignmentsService.setLegTransportMode` / `setIncomingLegTransportMode` and `DaysService.setDefaultTransportMode` respectively, with the same `?? null` coalescing, the same trip-access / `day_edit` permission checks, the same demo-user gate, and the same `assignment:updated` / `day:updated` broadcasts as the REST handlers.

Continues the per-segment transport feature line: the modes introduced in #1281 and the boundary-leg modes shipped in #1890. This PR adds the MCP surface for setting those same modes (the REST + UI paths already exist); it introduces no new user-facing behaviour, only a programmatic way to reach the existing setters.

## Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update

## Approach

The tools live on the existing `AssignmentsMcp` and `DaysMcp` controllers next to their siblings and follow the same shape exactly: inline Zod `inputSchema`, `TOOL_ANNOTATIONS_WRITE`, and the declarative access markers (`{ group: 'places', mode: 'write' }` for the leg tool, `{ group: 'trips', mode: 'write' }` for the day tool). `transport_mode` stays a loose nullable string, matching `assignmentTransportRequestSchema` / `dayTransportRequestSchema` in `@trek/shared` – the same source of truth the REST routes derive from, including `direction: z.enum(['outgoing','incoming']).default('outgoing')`.

The `set_leg_transport_mode` description states plainly that `"incoming"` only affects rendering when the leg's origin is not itself a place assignment, so an MCP caller does not silently set an inert mode on an ordinary place-to-place leg.

(Note for reviewers: the inline `z.*` `inputSchema` in `*.mcp.ts` is the established house convention across every MCP controller in this repo – it is not an omission of a `@trek/shared` contract.)

## What's in it

- `assignments.mcp.ts`: `set_leg_transport_mode` (branches on `direction`).
- `days.mcp.ts`: `set_day_default_transport_mode` (inline not-found result, matching this file's convention).
- tests (`tools-leg-transport-mode.test.ts`): outgoing and incoming set only their own column (column independence), `null` clears / inherits, `direction` defaults to outgoing, day-default returns its populated `assignments` array, plus cross-trip rejection, not-found, non-member access denial, and demo-user block for both tools.

Scoped to the MCP server surface; no schema, API, or i18n changes. Server typecheck clean, lint clean, the full `tests/unit/mcp` suite passes (756 tests).

## Checklist
- [x] I have added/updated tests that prove my feature works

### DIFF
--- a/server/src/nest/assignments/assignments.mcp.ts
+++ b/server/src/nest/assignments/assignments.mcp.ts
@@ -117,7 +117,7 @@ export class AssignmentsMcp {
 
   @Tool({
     name: 'set_leg_transport_mode',
-    description: 'Set the travel mode of a route leg for a place assignment. Use direction "outgoing" (default) for the common case: the leg leaving this stop toward the next. Use direction "incoming" ONLY when this stop\'s arriving leg originates from something that is not itself a place assignment (e.g. a flight/train booking arrival, or a morning hotel departure) – setting "incoming" on an ordinary place-to-place leg is stored but has no effect on route rendering, because it targets a column that is only read for non-place origins. transport_mode is a route profile key (e.g. "driving", "walking", "cycling", "transit"); null clears it so the leg inherits the day default.',
+    description: 'Set the travel mode of a route leg for a place assignment. Use direction "outgoing" (default) for the common case: the leg leaving this stop toward the next. Use direction "incoming" ONLY when this stop\'s arriving leg originates from something that is not itself a place assignment (e.g. a flight/train booking arrival, or a morning hotel departure) – setting "incoming" on an ordinary place-to-place leg is stored but has no effect on route rendering, because it targets a column that is only read for non-place origins. transport_mode is a route profile key: "driving", "walking", "cycling", or a plugin profile written as "plugin:<pluginId>/<profileId>". Any other value is stored but drawn as a driving route. null clears it so the leg inherits the day default.',
     inputSchema: {
       tripId: z.number().int().positive(),
       assignmentId: z.number().int().positive(),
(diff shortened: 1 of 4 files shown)


mauriceboe:
Thanks — this is the shape a new MCP tool should have: same service methods, same `day_edit` check, same cross-trip join, same broadcast as the REST routes, and the `direction: "incoming"` caveat spelled out for the caller instead of left as a surprise.

I pushed two commits before merging:

- `"transit"` isn't a route profile key. `RouteCalculator` knows `driving`, `walking`, `cycling` and `plugin:<id>/<profile>`, and maps anything else to driving when it draws the route — so an agent taking the example at its word got a success, a car route and a car icon. Both descriptions now name the real keys and say what happens to anything else, and the incoming-leg test uses `walking`. While in there, the outgoing test seeds `incoming_leg_transport_mode` first, so its "must not touch incoming" assertion proves something the factory didn't already give it.
- Added both tools to `wiki/MCP-Tools-and-Resources.md`; every other tool on those controllers has a row.

Issue #1372, opened 2026-06-29
[BUG] Google map link missing accomodations/hotels in route


starsoccer:
### Pre-flight checklist

- [x] I have searched [existing issues](https://github.com/mauriceboe/TREK/issues) and this bug has not been reported yet
- [x] I am running the latest available version of TREK
- [x] I have read the [Troubleshooting guide](https://github.com/mauriceboe/TREK/wiki/Troubleshooting) and my issue is not covered there

### TREK version

3.1.3

### Describe the bug

When clicking the google map link on a day, the linked route/map does not include starting/ending accommodations/hotels

### Steps to reproduce

1. Go to a day with a starting/ending hotel
2. Click the day and click the G for google maps
3. Notice how map/route does not include starting hotel/ending hotel

### Expected behavior

Starting/ending hotel should be included

### Deployment method

Docker (standalone)

### Accessing TREK from

Desktop browser

### Relevant logs or error output

```shell

```

mauriceboe:
Fixed — the "open in Google Maps" route now includes your accommodation/hotel stops, matching the route drawn on the map. Coming in **v3.2.0**.

PR #1932, opened 2026-08-14
feat(planner): open places and day routes in CoMaps

subdee:
## Description

Adds optional CoMaps integration for places and day routes. A place with coordinates
gains a "CoMaps" entry in its navigation menu; a day's route toolbar gains a button
that opens that day's stops in CoMaps for offline navigation, in the day's own
travel mode.

CoMaps offers two deep links that trade against each other: `route` builds real
turn-by-turn navigation in a given travel mode but holds only a start and a
destination, while `map` takes any number of named pins but routes nothing. So
`generateCoMapsUrl` (`client/src/components/Map/RouteCalculator.ts:119`) sends a
two-stop day as a route — nothing is lost, mode included — and a longer day as pins,
rather than quietly dropping the stops in between. A day that needs its full
itinerary as one navigable track already has the GPX export.

Links use the `https://comaps.at/…` form rather than the `cm://` scheme CoMaps also
answers to. Both open the app when it is installed, but only https has somewhere to
land when it is not, which is the fallback the discussion asked for — a PWA cannot
detect an installed Android app.

Layers touched: `client/src/components/Planner/placeCoMaps.ts` (new URL builder,
mirroring `placeOpenStreetMap.ts`); `placeNavigation.ts:90` (new `comaps` target);
`RouteCalculator.ts:119` (`generateCoMapsUrl`); `DayPlanSidebar.tsx:2478` and the
mobile surfaces `MDaySheet.tsx:328` / `MPlanTimeline.tsx:189` via
`dayRoute.ts`; `planner.openCoMaps` and `mobileTrip.coMaps` in all 23 locales.

The hotel-bookend rules behind the existing Google Maps export (#1372/#1465) were
inlined in the desktop sidebar and duplicated in the mobile lib. Both now read one
`dayExportStops`, so the Google and CoMaps links cannot disagree about which hotel
legs are real. This is why the `FE-PLANNER-DAYPLAN-168` expectation changed — the
shared stop list now carries place names, which the CoMaps links use as pin labels.

`placeNavigation.test.ts` carried a stale `// @vitest-environment jsdom` pragma that
broke every test in the file against the project's custom jsdom environment; removing
the line was necessary for the new test there to run.

The GPX day export (proposal 3 in the discussion) is deliberately out of scope — it
already shipped separately.

Addresses discussion #1904

## Type of Change
- [x] New feature

## Checklist
- [x] I have added/updated tests that prove my feature works


mauriceboe:
@subdee conflicts: client/src/components/Planner/DayPlanSidebar.tsx

mauriceboe (approved):
looks good!

PR #1970, opened 2026-08-18
fix(llm-parse): parse a stringified reservations field from the Anthropic tool response

### LINKED ISSUE #1968: [BUG] AI booking import with Anthropic silently returns no reservations when the model sends tool input as a JSON string
### Pre-flight checklist

### TREK version

3.4.1

### Describe the bug

With the AI Parsing addon set to the **Anthropic** provider, an import sometimes reports that no reservations were found even though the document contains one and the model did extract it. There is no error and nothing in the logs — the import simply comes back empty, indistinguishable from a document with no booking in it.

`server/src/nest/llm-parse/clients/anthropic.client.ts:82-84` accepts the tool result only when it is already an array:

```ts
const toolUse = data.content?.find(b => b.type === 'tool_use' && b.name === TOOL_NAME);
const reservations = toolUse?.input?.reservations;
return Array.isArray(reservations) ? (reservations as Record<string, unknown>[]) : [];
```

Despite the array-typed `input_schema` and the forced `tool_choice`, Claude sometimes returns `reservations` as a JSON-encoded **string** rather than an array. The ternary then falls through to `[]` and a good extraction is silently discarded. Two shapes have shown up in practice:

- a stringified array — `"[{\"@type\":\"LodgingReservation\", …}]"`
- a stringified, double-wrapped object — `"{\"reservations\":[{…}]}"`

Because it depends on how the model happens to serialise its tool call, the same document can import fine on one attempt and come back empty on the next.

### Steps to reproduce

1. Admin → Addons → enable **AI Parsing**; set provider **Anthropic** with an API key and a model.
2. Open a trip → **Bookings** → **Import from file**.
3. Upload a booking confirmation KDE Itinerary cannot parse (e.g. a plain-text email), so the AI fallback runs.
4. Repeat the same upload a few times. On some attempts the review step reports that no reservations could be extracted.

To be upfront: the trigger is the model's serialisation choice, so this is intermittent rather than deterministic — it can't be forced from the UI. The code path above discards the result whenever it does occur.

### Expected behavior

The extracted reservations are used whether the provider returns `reservations` as an array or as a JSON-encoded string. A field that genuinely cannot be interpreted should surface as a parse failure, rather than as an empty but successful-looking import.

### Deployment method

Docker Compose

### Host OS

Unraid 7.3.2

### Accessing TREK from

Desktop browser

### Browser (if applicable)

Chrome

```shell
Nothing is logged. The extracted text is written to the debug log as normal, and then
the import ends with "No reservations could be extracted from the uploaded files."
— identical to what a document containing no booking produces.
```

### Additional context

Happy to open a PR — I have a fix with tests covering both observed shapes, plus a case pinning that an genuinely unparseable value still yields `[]` rather than the fix over-reaching.

michael-bohr:
## Description

The Anthropic client accepts the tool result only when it is already an array, so a
`reservations` field the model returned as a JSON-encoded string is dropped and the
import reports nothing found — with no error and nothing logged.

`extract()` now parses a string value before the array check, and unwraps the
`{reservations:[...]}` object when the model double-wraps it. A value that still cannot
be interpreted returns `[]` exactly as before, so the change can't widen what counts as
a valid result.

Three tests: the two shapes observed in practice fail without the fix, and a third pins
that an unparseable value still yields `[]` — that one passes either way on purpose,
guarding the fix from over-reaching rather than pinning the bug.

Closes #1968

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Checklist
- [ ] I have updated documentation if needed *(no user-facing behaviour to document — the
      import simply stops losing results)*

### Test plan

- `npx vitest run tests/unit/nest/llm-parse/clients.test.ts` — 19 passed
- `npm run typecheck --workspace=server` — clean
- Reverting the client change fails the two new stringified cases and leaves the guard
  case passing, as intended

### DIFF
--- a/server/src/nest/llm-parse/clients/anthropic.client.ts
+++ b/server/src/nest/llm-parse/clients/anthropic.client.ts
@@ -80,7 +80,21 @@ export class AnthropicClient implements LlmExtractionClient {
     }
 
     const toolUse = data.content?.find(b => b.type === 'tool_use' && b.name === TOOL_NAME);
-    const reservations = toolUse?.input?.reservations;
+    let reservations = toolUse?.input?.reservations;
+    // The model is expected to fill `reservations` as a native array — that's what
+    // forcing tool_choice is for — but it occasionally stringifies the field instead,
+    // sometimes double-wrapped as `{"reservations": [...]}`. Parse defensively rather
+    // than letting the Array.isArray check below silently drop a good extraction to [].
+    if (typeof reservations === 'string') {
+      try {
+        const parsed: unknown = JSON.parse(reservations);
+        reservations = Array.isArray(parsed)
+          ? parsed
+          : (parsed as { reservations?: unknown } | null)?.reservations;
+      } catch {
+        reservations = undefined;
+      }
+    }
     return Array.isArray(reservations) ? (reservations as Record<string, unknown>[]) : [];
   }
 }
--- a/server/tests/unit/nest/llm-parse/clients.test.ts
+++ b/server/tests/unit/nest/llm-parse/clients.test.ts
@@ -226,4 +226,33 @@ describe('AnthropicClient', () => {
     const blocks = body.messages[0].content;
     expect(blocks.some((b: any) => b.type === 'document' && b.source.type === 'base64')).toBe(true);
   });
+
+  it('parses a stringified reservations array instead of dropping it to []', async () => {
+    // Observed in live use: despite the array-typed input_schema and a forced
+    // tool_choice, the model sometimes stringifies the field.
+    mockFetch(() =>
+      jsonResponse({
+        stop_reason: 'tool_use',
+        content: [{ type: 'tool_use', name: 'emit_reservations', input: { reservations: '[{"@type":"LodgingReservation"}]' } }],
+      }),
+    );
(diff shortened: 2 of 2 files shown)


mauriceboe:
Thank you for this, and sorry it sat here without a word.

I hit the same bug while working through the import reports and fixed it in #1988 before I had properly looked at your branch, which was already open and already correct. That is on me. Your diagnosis was right, your patch worked, and the tests you wrote cover the two shapes that actually occur.

What went in is the same fix moved one level up: the unwrapping now lives in `toReservationList()` in `lenient-json.ts` and both the Anthropic client and the OpenAI-compatible one call it, so Ollama and friends get it too. It parses through `parseLenientJson`, which also handles a value the model wrapped in a ```json fence or wrote as a JS object literal. Your three test cases are in `clients.test.ts` more or less verbatim.

So there is genuinely nothing left for this branch to add, and rebasing it would produce an empty diff. Closing it for that reason only.

For the record, since #1988 credits you as the reporter and that undersells it: you filed the issue, found the exact lines, and shipped a working patch with tests within two minutes of the report. Thanks — please keep sending these.

PR #1971, opened 2026-08-18
fix(booking-import): geocode imported transport endpoints during preview

### LINKED ISSUE #1969: [BUG] Imported transport endpoints are silently dropped when the extractor supplies no coordinates
### Pre-flight checklist

### TREK version

3.4.1

### Describe the bug

Importing a train, bus or rental-car confirmation shows the pick-up and drop-off in the review step's From→To summary, but once saved the endpoints are gone from the booking — no map pins, no route, and no error. The reservation itself saves correctly; only the endpoints vanish. Checking the database after a save, `reservation_endpoints` had no rows at all for the trip.

The cause is that nothing geocodes these endpoints on the path the app actually uses.

1. `kitinerary-mapper.ts` emits named endpoints with `lat: null, lng: null` for non-airport locations (`:193-194` trains, `:300-301` rental cars) — deliberately, with comments saying `confirm()` geocodes them later.
2. `BookingImportService.confirm()` does contain exactly that geocoding (`booking-import.service.ts:181-197`) — **but nothing calls `confirm()`**. Its client wrapper `reservationsApi.importBookingConfirm` (`client/src/api/client.ts`) has no callers outside tests. Both the desktop planner (`useTripPlanner.ts`) and mobile (`MReservationSheet.tsx`) take the review flow instead: `preview()` → open each parsed item in the normal edit modal → save through the generic `handleSaveReservation` / `handleSaveTransport`. So that geocoding never runs.
3. Nothing else fills the gap. `TransportModal.tsx:69-72` passes the null coordinates straight through (and hardcodes `address: null`), and the save handler is a passthrough.
4. `ReservationsService.saveEndpoints()` (`server/src/nest/reservations/reservations.service.ts:332`) then filters out endpoints without coordinates before inserting, because `reservation_endpoints.lat`/`lng` are `NOT NULL`. That filter is a sensible, documented guard — it is just where the loss becomes visible.

Net effect: an endpoint the extractor named but could not geo-locate is dropped on save, with nothing shown to the user. Flights are unaffected, since the mapper resolves airports from its own airport table.

### Steps to reproduce

1. Create a trip with start and end dates.
2. Open **Bookings** → **Import from file** and upload a car rental confirmation whose pick-up location has a street address and a name that isn't a recognisable place on its own (e.g. a desk name like `Curbside Pickup Counter 7`).
3. In the review step, note that the pick-up/drop-off appear in the From→To summary.
4. Click **Add** to save the booking.
5. Open the saved booking — the endpoints are missing, and no pin or route appears for them.

### Expected behavior

An imported endpoint that arrives without coordinates should be geocoded — by name, falling back to the address on the same confirmation — so that it survives the save. If it genuinely cannot be resolved, the user should be told, rather than having it disappear silently.

### Deployment method

Docker Compose

### Host OS

Unraid 7.3.2

### Accessing TREK from

Desktop browser

### Browser (if applicable)

Chrome

```shell
Nothing is logged — the booking saves successfully and the endpoint loss is silent.
Confirmed against the database after a save:

  SELECT e.* FROM reservation_endpoints e
    JOIN reservations r ON r.id = e.reservation_id
   WHERE r.trip_id = <trip>;
  -> 0 rows

while the reservation row itself is present with its title and confirmation number.
```

### Additional context

Happy to open a PR. The fix I have moves the geocoding into `preview()` — the path the clients actually take — trying the endpoint name first and then the address from the same confirmation, on the throttled `'background'` Nominatim lane. `MapsService.searchNominatim`'s own docstring (`maps.service.ts:574-583`) already describes booking-import's endpoint loop as that lane's intended caller, so the infrastructure is there; the loop just sits in the method nothing reaches. Endpoints that still cannot be resolved are kept rather than filtered, so they stay editable in the review form, and they now raise a warning instead of disappearing without a word.

It deliberately does **not** touch `confirm()` — whether that orphaned path should be removed or rewired seems like a separate call for you to make, and its own geocoding is a no-op on already-coordinated endpoints, so the two can't conflict.

michael-bohr:
## Description

An imported train, bus or rental-car endpoint arrives from extraction with a name but no
coordinates, and nothing geocodes it: the mapper defers to `confirm()`, but the clients
never call `confirm()` — they open each parsed item in the normal edit modal and save
through `ReservationsService`, whose `saveEndpoints()` drops any endpoint still missing
coordinates because `reservation_endpoints.lat`/`lng` are `NOT NULL`. The endpoint shows
in the review form's From→To and is silently gone from the saved booking.

This moves the geocoding into `preview()`, which is the path the clients actually take.
It tries the endpoint name first, then the address from the same confirmation, on the
throttled `'background'` Nominatim lane — `MapsService.searchNominatim`'s docstring
already names booking-import's endpoint loop as that lane's intended caller, so this is
the load pattern it was built for.

Two deliberate differences from the version in `confirm()`:

- **Unresolvable endpoints are kept, not filtered.** They stay visible in the review
  form's From→To so the location can be set by hand before saving.
- **They raise a warning** naming the endpoint, instead of disappearing without a word.

`confirm()` is left exactly as-is. Whether that orphaned path should be removed or
rewired felt like a separate decision, and its own geocoding is a no-op on
already-coordinated endpoints, so the two can't conflict.

The only change outside the server is a one-line comment correction in the shared schema
(it said `confirm()` geocodes these; it's now `preview()`).

Closes #1969

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Checklist
- [x] I have updated documentation if needed *(two stale code comments corrected; no
      user-facing docs affected)*

### Test plan

- `npx vitest run tests/unit/nest/booking-import/booking-import.service.test.ts tests/unit/services/kitineraryMapper.test.ts` — 15 passed
- `npm run typecheck --workspace=server` — clean
- ESLint on the touched files: 19 warnings, identical to the pre-change baseline for the
  same files — no new warnings, no new `any`
- With the `preview()` change reverted, 5 of the 6 new tests fail; the sixth ("does not
  re-query an endpoint that already has coordinates") passes either way by design, since
  it guards against the fix over-reaching
- Also verified end-to-end on a live instance: before the change, importing a rental-car
  confirmation left `reservation_endpoints` with zero rows for the trip while the
  reservation row itself saved fine

### DIFF
--- a/server/src/nest/booking-import/booking-import.service.ts
+++ b/server/src/nest/booking-import/booking-import.service.ts
@@ -111,8 +111,9 @@ export class BookingImportService {
         const { items, warnings } = mapReservations(kiItems, file.originalname);
         // LLM extraction is less certain than kitinerary — always flag for review.
         if (aiUsed) for (const it of items) it.needs_review = true;
-        allItems.push(...items);
         allWarnings.push(...warnings);
+        allWarnings.push(...(await this.geocodeEndpoints(items, file.originalname)));
+        allItems.push(...items);
       }
 
       // Report per-file progress so a background import can drive a live widget.
@@ -120,6 +121,56 @@ export class BookingImportService {
     }
 
     return { items: allItems, warnings: allWarnings, files: fileReports };
+  }
+
+  /**
+   * Geocode transport endpoints (stations/stops/terminals/rental desks) that arrived
+   * without coords, so the route draws and map pins appear. The LLM and kitinerary
+   * rarely supply geo for anything but airports.
+   *
+   * This belongs to preview() because preview is the path the clients actually take:
+   * each parsed item is opened in the normal edit modal and saved through
+   * ReservationsService, never through confirm(). An endpoint still lacking coords at
+   * save time is dropped by saveEndpoints() — reservation_endpoints.lat/lng are NOT
+   * NULL — so it would vanish from the booking with nothing said. Hence the warning
+   * instead of a silent skip.
+   *
+   * Unresolvable endpoints are kept, not filtered: they still appear in the review
+   * form's From→To, where the user can set the location by hand before saving.
+   */
+  private async geocodeEndpoints(items: ParsedBookingItem[], fileName: string): Promise<string[]> {
+    const warnings: string[] = [];
+    for (const item of items) {
+      if (!Array.isArray(item.endpoints)) continue;
+      for (const ep of item.endpoints) {
+        if (ep.lat == null || ep.lng == null) {
+          // Name first, then the address: a rental desk's name is often unresolvable
+          // ("Curbside Counter 7") while the address on the same confirmation is exact.
+          for (const query of [ep.name, ep.address]) {
+            if (!query) continue;
+            try {
+              // 'background': an import geocodes every uncoordinated endpoint in one
+              // loop, and must yield to anyone typing in the interactive place search.
+              const hit = (await this.maps.searchNominatim(query, undefined, 'background'))[0];
+              if (hit?.lat != null && hit?.lng != null) {
+                ep.lat = hit.lat;
+                ep.lng = hit.lng;
+                break;
+              }
+            } catch {
+              // Geocoding failure is non-fatal — the booking still imports.
+            }
+          }
+          if (ep.lat == null || ep.lng == null) {
+            warnings.push(`${fileName}: could not locate "${ep.name}" — set it manually on the booking, or it won't be saved`);
+          }
+        }
+        // Transient: neither reservation_endpoints nor the wire contract
+        // (bookingImportEndpointSchema) carries an address.
+        delete ep.address;
+      }
+    }
+    return warnings;
   }
 
   /**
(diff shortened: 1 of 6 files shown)


mauriceboe:
Same story as #1970, and the same apology: this was open and correct while I fixed the identical bug in #1993 without checking here first.

The version that landed moves the lookup out of `confirm()` into `preview()` the way yours does, keeps unresolvable endpoints instead of dropping them, and raises a warning per endpoint. It also does three things yours did not, and one of them matters to your change specifically: `BackgroundTasksWidget` only rendered import warnings in the "no items found" branch, so the warnings your patch raises would have been invisible on exactly the imports that raise them. The other two are a helper `confirm()` also calls, so the two paths cannot drift apart again, and a per-preview cache for the Nominatim lane.

One deliberate difference worth explaining, because it goes the other way: your ladder queries the bare endpoint name first and takes the first hit. For a rental desk with an address that is fine, but for trains, buses and ferries the name is all there is, and "Hauptbahnhof" or "Central Station" resolves to *a* place somewhere in the world. Writing that coordinate is worse than writing none, because it gets pinned on the map and drawn into the route with nothing marking it as a guess. So the merged version tries name plus the booking's location first and leaves the bare name for last.

Your branch does hold one thing dev still lacks, and you were right about it: you carry the pick-up and drop-off addresses separately on each endpoint, while dev derives both from the pick-up. On a one-way rental that means the drop-off desk is currently geocoded against the pick-up's address. I will fix that separately and I would not have spotted it without this branch.

Closing as superseded rather than because anything is wrong with it. Thanks for the report, the patch and the tests.

PR #1996, opened 2026-08-20
feat(budget): add "Fuel" and "Parking" expense categories

MikeDabrowski:
## Summary
- Adds two new fixed budget categories: `fuel` ("Fuel") and `parkings` ("Parking"), alongside the existing set (Accommodation, Food & drink, Groceries, Transport, Flights, Activities, Sightseeing, Shopping, Fees & tickets, Health, Tips, Other).
- Icon/color metadata added for both (reusing `Fuel` and `ParkingCircle` from lucide-react), plus legacy free-text aliases (gas/petrol→fuel, parking→parkings) for existing/imported expense rows.
- Label strings added for all 23 supported locales.

## Test plan
- Added `shared/src/budget/budget.schema.spec.ts` coverage confirming `fuel`/`parkings` are in `COST_CATEGORIES`.
- Added `client/src/components/Budget/costsCategories.test.ts` covering category metadata, list ordering, and legacy-alias resolution (gas/petrol→fuel, parking→parkings).
- Manually verified in a running instance: new categories appear in the "Add expense" category picker and can be assigned to expenses.


mauriceboe:
Thanks for this — the shape of it is right, and doing all 23 locales properly rather than leaving 22 English placeholders is appreciated. Fuel in particular is a gap I have wanted closed for a while: road-trip petrol has had nowhere sensible to go.

I pushed one commit onto your branch rather than sending you round again for two small things.

**The key is now `parking`, not `parkings`.** The reservation type of the same name is spelled `parking` everywhere else in the codebase, and there is a place where the difference is visible rather than academic: the public share page prints the stored category verbatim, with no lookup and no translation, so a share link would have read "parkings" in every language. The translated values are unchanged, only the key moved.

**A parking booking now files its expense under Parking.** `RESERVATION_TYPE_TO_COST_CATEGORY` still had `parking: 'transport'` — I set that line myself in #1642, when there was no parking category to send it to. Left as it was, your new bucket would sit empty next to a Transport row holding exactly the costs that belong in it, while a manually typed "parking" resolved into the bucket. Both paths agree now, with a test pinning it.

Also in there: `catMeta('Parking')` and `catMeta('car park')` resolve too, and one Indonesian string now matches the casing of its neighbours.

Existing expenses are untouched — nothing rewrites a category that is already stored, so anything filed under Transport stays there.

The workflows on this PR are waiting on approval, which is why the only green check is the target-branch one. I have run them locally: shared and client typecheck, i18n file and key parity pass, and the 257 Budget tests plus the reservations and booking-import suites are green. I will approve the run so CI says so too.

mauriceboe (approved):
Green across all nine checks. Merging.

MikeDabrowski:
Great! Thank you for the detailed explanation - I wasn't expecting that and
I much appreciate it. Makes perfect sense.

PR #2015, opened 2026-08-22
feat(trips): hide collab tab when trip has no companions

syyangv:
Hide the collaboration tab on trips where the user has no travel companions (solo trips), matching the sharing element visibility rules.


syyangv:
Opened on fork by accident, closing.

Issue #1541, opened 2026-07-13
[BUG] Active filter not showing on places/activities after changing menus


braceyourself07:
### Pre-flight checklist

- [x] I have searched [existing issues](https://github.com/mauriceboe/TREK/issues) and this bug has not been reported yet
- [x] I am running the latest available version of TREK
- [x] I have read the [Troubleshooting guide](https://github.com/mauriceboe/TREK/wiki/Troubleshooting) and my issue is not covered there

### TREK version

3.3.0

### Describe the bug

When using the filter on the Places/Activities list, it will work.  But then if you change to another menu like "Book" then return to the map, the filter reverts to "All Categories" with nothing marked but the map is still filtered by what was previously selected. The only way I can fix it is to reselect and deselect something so it truly goes back to showing All Categories.

### Steps to reproduce

1. On "Plan" filter list of Activities/Places by something (e.g. Hotel)
2. Go to "Book" then return to "Plan"
3. The filter will still be active although it says "All Categories"
4. The only way to turn off the filter is to select any filter and then deselect

### Expected behavior

I would expect the active filter to still show upon returning to the "Plan" tab after visiting a different tab. 

### Deployment method

Docker Compose

### Host OS

Debian 13

### Accessing TREK from

Desktop browser

### Browser (if applicable)

Chrome Version 150.0.7871.115

### Relevant logs or error output

```shell

```

### Screenshots

<img width="1484" height="651" alt="Image" src="https://github.com/user-attachments/assets/4a9d11fc-2782-4d95-8a10-3025574da81d" />
List filtered by Hotel

<img width="1786" height="895" alt="Image" src="https://github.com/user-attachments/assets/79169c16-1a6b-40bf-ae16-7e5d8c75945f" />
After going to the Book tab and returning to Plan, only Hotels show while the filter says "All Categories"

<img width="1839" height="794" alt="Image" src="https://github.com/user-attachments/assets/b0f84300-575e-47ce-809a-e69af10764f4" />
All Categories only truly shows up only after toggling any filter on/off.

PR #2025, opened 2026-08-23
feat(planner): planned filter follows the selected day; day header click toggles selection

Cartoone9:
Follow-up to discussion #2024 (items 1 and 2), offered as code in case it helps — happy to adjust or split if you'd prefer a different shape. Based on `dev` since that's where 4.0 work lives.

## 1. `planned` filter follows the selected day

With a day selected, the **Assigned/planned** places filter now shows only that day's stops — the same way the other filters follow the selection. With no day selected, behavior is unchanged: the whole plan stays on the map (including collapsed days, per the existing #1541 behavior). `unplanned` deliberately keeps using the whole-trip assigned set — a place assigned to any day is not unplanned.

| Filter | No day selected | Day selected |
|---|---|---|
| planned (before) | whole plan | whole plan |
| planned (after) | whole plan | **selected day only** |
| unplanned | unchanged | unchanged |

## 2. Clicking the selected day's header deselects it

A second click on the already-selected day header now clears the selection and closes the day detail panel — identical to pressing the panel's ×. The `onDayDetail` prop type widens to `(day: Day | null) => void` to carry the close.

## Tests

- `FE-TP-HOOK-111`: planned filter scoped with a day selected, whole plan again at `selectedDayId: null`
- `FE-PLANNER-DAYPLAN-209`: second click on the selected header calls `onSelectDay(null)` + `onDayDetail(null)`

Both touched suites pass (344/344).

Item 3 of #2024 (visually distinguishing unassigned pins) is not included — it seems to want a product decision on option vs. default, so I left it to the discussion.

### DIFF
--- a/client/src/components/Planner/DayPlanSidebar.test.tsx
+++ b/client/src/components/Planner/DayPlanSidebar.test.tsx
@@ -370,6 +370,17 @@ describe('DayPlanSidebar', () => {
     expect(onSelectDay).toHaveBeenCalledWith(10)
   })
 
+  it('FE-PLANNER-DAYPLAN-209: clicking the selected day header again deselects it', async () => {
+    const user = userEvent.setup()
+    const day = buildDay({ id: 10, date: '2025-06-01', title: 'My Day' })
+    const onSelectDay = vi.fn()
+    const onDayDetail = vi.fn()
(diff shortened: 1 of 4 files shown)


mauriceboe (approved):
Thanks for both the write-up and the patch. Clean, tested, and it makes `planned` behave like every other filter on the map instead of being the one exception the collapsed-day declutter skipped. The day-header toggle is an obvious win too.

Merging for 4.0.0.

PR #2035, opened 2026-08-25
Feat/switch plugin version

jubnl:
## Description
Add a way to install an older version of a plugin


mauriceboe (approved):
Read through it all and ran the gates locally against current dev: typecheck, typecheck:tests, the plugin suites (68 files / 940 tests), both panel suites, i18n parity and lint all green. The 13 new keys are present and actually translated in all 23 locales, and resume-updates inherits the admin guard from the controller.

Two things I pushed onto the branch rather than sending back:

- A test for the update hold. Every fixture left `trek` unset, so newest-published and newest-compatible were the same row and the assertions passed either way. Swapping resolveVersion() for versions[0] kept the file green, so the distinction was unpinned.
- The merge with dev. The Sonar sweep had rewritten the same "Update all" line, and taking its side would have reintroduced the bug your arrow function guards against: forEach passes the index, which lands in runUpdate's version parameter. Kept both sides, and gave the seven buttons this branch adds the same explicit type.

Nice work, thanks.

PR #2055, opened 2026-08-27
Improve mobile itinerary focus and plan discovery

cyzbcf-beep:
## Summary
- select the relevant itinerary day when the mobile trip shell opens: today, the next future day, or the final completed day
- keep automatic selection to a single initial seed so later user choices are not overwritten
- calculate UP NEXT from the effective assignment time (`assignment_time` before `place_time`) and render that same display time
- hide stale UP NEXT content after today's timed plan is exhausted and for all historical dates
- adapt the behavior to the current `MTripShell` / `MPlanTimeline` architecture without restoring the legacy mobile planner implementation or changing map coordinates

## Validation
- affected mobile/date tests: 145 passed
- client TypeScript typecheck: passed
- changed-file ESLint and page-pattern lint: passed
- client production build and PWA generation: passed
- `git diff --check` and added-line security scan: passed
- final independent code/security review: passed with no blocking logic or security findings
- local full Windows client run: 613/617 files passed; 12,651 tests passed, with 34 failures in four unrelated UI/translation test files (not classified as pre-existing without a clean-base comparison)

## Compatibility
- rebuilt as a focused eight-file integration on current `dev` (`d194d2cf`), rather than cherry-picking the TREK 3.4.1 implementation
- preserves the current independent mobile shell, plan timeline, map area, desktop layout, PWA architecture, and existing itinerary data
- does not fabricate missing coordinates or change server/database contracts

### DIFF
--- a/client/src/components/Planner/today.test.ts
+++ b/client/src/components/Planner/today.test.ts
@@ -1,5 +1,5 @@
 import { describe, it, expect } from 'vitest'
-import { findTodayDayId, localToday } from './today'
+import { findRelevantDayId, findTodayDayId, localToday } from './today'
 
 describe('localToday', () => {
   it('FE-TODAY-001: reads the date off the local clock, not UTC', () => {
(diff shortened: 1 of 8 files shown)


mauriceboe:
Thanks for this — the UP NEXT part especially is a real fix, and I've carried it forward in #2057 along with two of the other pieces:

- UP NEXT stays away on a day that has passed and once today's timed plan has run out
- the seed guard treating an already-selected day as seeded, so a later deselect isn't fought
- seeding the next day still ahead when today falls into a gap in the dates

Two things I left out, so you know why rather than wondering:

**Opening a finished trip on its last day.** Coming back to a trip that's over, you tend to read it from the start rather than the end, so I kept the existing first-day fallback and limited the new helper to days that are still ahead.

**`assignment_time || place_time` in `findUpNext`.** That override is already resolved before it reaches the timeline model — the server projects `COALESCE(da.assignment_time, p.place_time)` into the embedded place, and `mergeAssignmentPlace` in `placesSlice.ts` keeps that true on the client. The comment there notes that two spellings of that rule is how those paths drifted apart once already, so I'd rather not add a third. If you have a case where an override genuinely doesn't come through, please open an issue with it — the fix would belong in `mergeAssignmentPlace`, not in the timeline model.

Closing this in favour of #2057. Thanks for digging into the mobile plan.

PR #2102, opened 2026-08-28
fix(atlas): retile the map instead of rebuilding it when the CARTO ke…

### LINKED ISSUE #2097: [BUG] App crashes with "can't access property "filter", k is undefined" after a CARTO API key is saved
### Pre-flight checklist

### TREK version

4.0.0

### Describe the bug

After I save a CARTO API key in Settings → Map, parts of the app crash. An error boundary shows:

> can't access property "filter", k is undefined

The crash does not happen every time. It happens at different places in the app. When I remove the key, the crash never happens.

I read the v4.0.0 client source. The key arrives late: `loadSettings()` fills the settings store after the first render. When the key arrives, `useTileUrl` (`client/src/hooks/useTileUrl.ts:13`) returns a new URL string with `?key=...`.

Atlas and JourneyMap use this URL as an effect dependency:

- `client/src/pages/atlas/useAtlas.ts:86` and `:447` — the effect removes the whole Leaflet map and builds it again.
- `client/src/components/Journey/JourneyMap.tsx:342` and `:385` — the same pattern, added by #2060.

During this teardown and rebuild, Leaflet event callbacks and third-party code (react-leaflet-cluster) still run against the old map. I think this race throws the minified `k.filter` error. The race also explains why the crash is intermittent. With no key, the URL never changes, so the map is never rebuilt.

### Steps to reproduce

1. Use the Leaflet map provider with the default OpenStreetMap tile template.
2. Save a CARTO API key in Settings → Map.
3. Open the Atlas, a Journey map, or a Collections map a few times.
4. Sometimes an error boundary appears with `can't access property "filter", k is undefined`.

### Expected behavior

The maps keep working when the key arrives. A key change must not rebuild the whole map, or the rebuild must be safe against the race.

### Deployment method

Kubernetes / Helm

### Host OS

OpenSUSE MicroOS (k3s cluster, Flux-managed container)

### Accessing TREK from

Desktop browser

### Browser (if applicable)

Firefox (crash message is Firefox wording), also seen in Chrome

```shell
server error are no helpful logs about this

Weather error: [TypeError: fetch failed] {
  [cause]: AggregateError [ETIMEDOUT]:
      at internalConnectMultiple (node:net:1339:18)
      at internalConnectMultiple (node:net:1415:5)
      at Timeout.internalConnectMultipleTimeout (node:net:1969:5)
      at listOnTimeout (node:internal/timers:637:11)
      at process.processTimers (node:internal/timers:571:7) {
    code: 'ETIMEDOUT',
    [errors]: [ [Error], [Error] ]
  }
}
[ERROR] 2026-08-28T08:32:44 GET /api/weather 500 258ms ip=
Weather error: [TypeError: fetch failed] {
  [cause]: AggregateError [ETIMEDOUT]:
      at internalConnectMultiple (node:net:1339:18)
      at internalConnectMultiple (node:net:1415:5)
      at Timeout.internalConnectMultipleTimeout (node:net:1969:5)
      at listOnTimeout (node:internal/timers:637:11)
      at process.processTimers (node:internal/timers:571:7) {
    code: 'ETIMEDOUT',
    [errors]: [ [Error], [Error] ]
  }
}
[ERROR] 2026-08-28T08:32:45 GET /api/weather 500 255ms ip=


and browser console are no any output.
```

### Screenshots

Image

Image

Image

### Additional context

- The CARTO key feature is new in v4.0.0 (#2054, fixed by #2060).
- Deployment is Docker image `mauriceboe/trek:4.0.0` behind Cloudflare, served from k3s.
- No fix on `dev` yet (I checked the commits after the v4.0.0 bump).
- Workaround: leave the key empty. The CARTO watermark returns, but the crash stops.

subdee:
## Description
`useTileUrl` returns a new string once `carto_api_key` reaches the settings store,
which happens after the first render. That string was a dependency of the effect
that *builds* the Leaflet map — `client/src/pages/atlas/useAtlas.ts:447`
(`}, [dark, loading, tileUrl]`), whose cleanup is `map.remove()` — so saving a
CARTO key made every Atlas load tear the map down and build a new one. With no key
the string never changes, which is why the crash only appears once a key is set.

Two things break in that window. Leaflet finishes a queued redraw against the
canvas renderer created at `useAtlas.ts:396` whose map has gone
(`Cannot read properties of undefined (reading 'save')` at `Canvas._clear` ←
`_redraw`). And the country layer never comes back: its effect (`useAtlas.ts:458`)
does not list `tileUrl`, so the rebuilt map shows tiles and nothing else — 12 of 12
Atlas loads with a key came back with no country canvas at all before this change.

`JourneyMap.tsx:342`/`:385` had the same shape, with `cartoKey`/`mapTileUrl` in the
dep arrays of the effects that draw every marker and track.

The tile template is a property of the tile layer, not of the map. Both maps now
keep refs to their `L.TileLayer`s, read the template through a ref inside the build
effect, and `setUrl()` it from an effect of its own. `setUrl` no-ops on an unchanged
URL, so the mount pass costs nothing.

Note: `react-leaflet-cluster` is not involved — the only `MarkerClusterGroup` is in
`MapView`, which takes the URL as a `<TileLayer url>` prop and is not rebuilt.

Closes #2097

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Checklist

### DIFF
--- a/client/src/components/Journey/JourneyMap.test.tsx
+++ b/client/src/components/Journey/JourneyMap.test.tsx
@@ -37,7 +37,7 @@ vi.mock('leaflet', () => {
   return {
     default: {
       map: vi.fn(() => mockMap),
-      tileLayer: vi.fn(() => ({ addTo: vi.fn() })),
+      tileLayer: vi.fn(() => ({ addTo: vi.fn(), setUrl: vi.fn() })),
       marker: vi.fn(() => mockMarker),
       polyline: vi.fn(() => { const line: any = { addTo: vi.fn(() => line), bindTooltip: vi.fn(() => line) }; return line }),
       divIcon: vi.fn(() => ({})),
@@ -45,7 +45,7 @@ vi.mock('leaflet', () => {
       layerGroup: vi.fn(() => ({ addLayer: vi.fn(), addTo: vi.fn(), remove: vi.fn() })),
     },
     map: vi.fn(() => mockMap),
-    tileLayer: vi.fn(() => ({ addTo: vi.fn() })),
+    tileLayer: vi.fn(() => ({ addTo: vi.fn(), setUrl: vi.fn() })),
     marker: vi.fn(() => mockMarker),
     polyline: vi.fn(() => { const line: any = { addTo: vi.fn(() => line), bindTooltip: vi.fn(() => line) }; return line }),
     divIcon: vi.fn(() => ({})),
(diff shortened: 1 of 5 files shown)


mauriceboe (approved):
lgtm

PR #2127, opened 2026-08-28
feat(plugins): let plugins expose MCP tools

RenzoBeux:
Reimplements #1752 against current `dev`, following your 28 Aug answers. Fresh PR as you asked. Thirteen commits, each green on its own.

## One place I went against your spec

You designed Q2 as `z.looseObject({}).check(...)`, required-only, and I built that first. A review pass showed it advertises constraints nothing enforces: a tool declaring an integer, an enum and `additionalProperties: false` accepted a string, an unlisted value and an extra property. The SDK and every client read a published `inputSchema` as "arguments are validated before the handler runs", so the model formats its call to a contract the runtime ignores.

The schema under `.meta()` is now built from the same normalised declaration and enforces it. Your objection to #1752's converter was that it flattened the advertisement into `unknown`; this does not, because `.meta()` still advertises the declaration byte for byte. What it adds is that the two cannot disagree: any keyword the builder cannot enforce (`oneOf`, unknown formats, non-primitive enum members) is refused by `parseManifest` at install time rather than advertised. No ajv, no new dependency.

That test also caught a real bug: `z.custom()` cannot be represented in JSON Schema, so the obvious way to build enums throws inside `tools/list` and takes out the surface for the whole session. Enums and consts are literal types.

If you would rather keep required-only, the revert is one commit and I will do it.

## Your three answers

**Q1.** Option A on `McpAttachOptions`, per session, with all four constraints: names reserved against every entry in `this.bound` rather than the session's grants, skip-never-throw per entry (both live throw sites), synchronous source, process-level sink. `mcp-transport.service.ts` is untouched, so the `mcp.tool_call` row for a plugin tool comes out of code that did not change.

**Q2.** Nothing to delete, the converter never existed on `dev`. See above.

**Q3.** One coarse `plugins:use`. Two defaults needed closing, not one: the client presets (five of six were "everything except deletes", duplicated desktop and mobile with nothing pinning them) and the DCR fallback in `oauth-sdk.provider.ts`, which handed `ALL_SCOPES` to any registration naming no scopes and pre-selected them at consent.

Both design changes and the three bugs are in too.

## Five other calls, push back on any

1. **`capabilityList` returns `string[]`**, so a parallel `mcpToolCapabilities(id)` carries this, re-running the manifest parser rather than trusting a blob written by an older one.
2. **`text-sanitize.ts` extended, not modified.** `stripEmoji` preserves newlines on purpose and ~15 render sites depend on it. The sibling also strips bidi overrides and U+2028/9, which are neither C0 controls nor `\n` and defeat both halves of the rule.
3. **A second i18n gate.** `client/tests/unit/i18n/parity.test.ts` diffs merged bundles, which do contain colon keys, so `oauth.scope.plugins:use.*` in en alone fails `client-coverage`. All 23 locales get everything.
4. **Caps you did not specify**: 8 tools per plugin, 32 overall, 32 content blocks and 64 KiB per whole result. Truncation is logged, since a silent one reads like a plugin that published nothing.
5. **`db:own` is not a side-effect grant** for `readOnlyHint`; counting the plugin's own SQLite would kill auto-approval for every legitimate read tool. Everything else is classified fail-closed and walked by a test, after a denylist version quietly let `db:create:trips` advertise itself read-only.

New Claude/Cursor/Windsurf/Zed clients do not request `plugins:use`, and existing tokens never carry it, so current sessions see no plugin tools until re-authorization.

## The tests you asked for

Integration, because the point is that nothing reaches the global filters. **MCP-P15**: a tool colliding with `list_trips` is skipped and the built-in still serves. **MCP-P16**: a throwing source degrades the surface instead of 500ing every initialize. **MCP-P17**: a plugin tool call writes the same audit row as a built-in. **MCP-P18** came out of writing P17, whose first draft minted a `trips:read` token and got no row because the access marker had correctly hidden the tool.

`mcp:tools` and the `mcpToolProvider` hook land in one commit: `requireTotalCoverage` makes a `HOOK_PERMISSION` key without a consumer a boot failure, and that runs in every `buildApp()`.

## Test plan

`typecheck`, `typecheck:tests`, `lint:check`, `check:plugin-facts`, `i18n-parity --strict`, the client bundle parity test and the plugin-sdk suite all pass. Coverage stays above every floor I touch (`src/nest/plugins/**` at 87.8/82.9/80.0/90.9 against 86/81/78/89). Roughly 120 new cases across the schema, enforcement, annotation, advertise, invalidation and preview paths.

My local suite has failures unrelated to this branch (backup/storage `EPERM: symlink` on Windows, client upload timeouts); they reproduce on a clean `dev`.


mauriceboe:
Merged.

Taking the Q2 deviation first, since you offered the revert: keep it. Advertising every keyword and enforcing none would have been a total drift, just a deliberate one, and your builder closes it for every node it walks. The revert would also have made two of the problems below worse, because both sit in the normaliser the spec asked for either way. What I did trim is the claim that the two cannot drift, since that was true only position by position.

Fixes pushed onto the branch:

`pattern` went straight into `new RegExp` and ran on the main thread against caller input with no timeout. `^(a+)+$` against 33 characters wedged a worker for 178 seconds in my test before I killed it, and that stops the whole server, not just MCP. It is an escape from the supervised child into the host event loop and it needs no malice, since a hand-written email pattern is the classic accident. Quantified groups that repeat or alternate inside are refused now, and the source is capped.

`isReadOnlyGrant` did not know `mcp:tools`, and holding that grant is what lets a plugin advertise at all, so `readOnlyHint` was structurally false for every plugin tool and every client's auto-approval was off. Your `db:own` reasoning applies to it exactly: it is a host entry point like `hook:*`.

The keyword allowlist only bound the nodes `nodeToZod` types, while `.meta()` advertises all of them. A `$ref` inside an object enum member, an `additionalProperties` subschema, `properties` under a string, anything beside an `enum`: all of it reached the advertisement unchecked. The allowlist now lives in `sanitiseNode`, the one walker that sees every node. Property names, `default` and `examples` were the strings that never met the sanitiser, and a key may carry newlines, so the collapse everything else goes through was bypassed there.

Smaller ones: a root `enum`/`const`/`nullable` produced a non-object schema the SDK cannot read, `exclusiveMinimum: true` (the draft-4 spelling) was dropped from the check but kept in the advertisement, `required` used `in` so a field named `toString` read as present on every call, `FORMAT_CHECKS` was a plain object so `format: "constructor"` passed the gate, and a shortened result block got no marker, which reads to the model like the whole answer. The wiki has `plugins:use` now; the scope count beside it was stale too.

Also: `mcp-tool-schema.test.ts` carried two raw NUL bytes, so git, GitHub and ripgrep all treated it as binary. No diff, no blame, no review view on the file that is the security control for the whole feature. Escaped, same runtime value.

Two left for you. `validateManifest` in the SDK is much more lenient than `parseMcpToolCapabilities`, so `trek-plugin validate` greens manifests the host refuses at install, and my changes widen that gap. The right shape is generating the caps and the keyword list into `host-facts.ts` rather than mirroring them by hand. And `mcpToolCapabilities()` has no test of its own; every test stubs it, so deleting the re-parse its own docblock argues for would stay green.

Good work on this one. The hook design, the manifest intersection and the audit seam all landed as specified, and the parts I changed are the ones a second pair of eyes is for.

PR #2130, opened 2026-08-29
refactor(places): one door onto the place-matching rule

michael-bohr:
## Description

Two consumers answer "is this candidate a place the trip already has?", and they spelled the
rule out separately. The bulk importers ask an in-memory set built for the trip
(`isPlaceDuplicate`); `findDuplicatePlace` asks the database, because its caller needs the
matched row's id rather than a yes/no.

The two had drifted. `findDuplicatePlace` falls back to a coordinate match for a **named**
candidate; `isPlaceDuplicate` deliberately never does, because widening the coordinate check
to named places merges the restaurant and the bar at the same address — a worse failure than
the duplicate it would prevent, as the comment there says.

That divergence is unreachable today. The only caller of `findDuplicatePlace` sits inside an
`isPlaceDuplicate(...)` guard that has already matched on name and returned, so the
coordinate branch never runs for a named candidate on that path. It arms the moment anything
calls it standalone — which is what the booking-import dedupe would have been.

So rather than patch one copy, the order now lives once in `@trek/shared` as
`placeMatchStrategies()`, and each consumer is a mechanical interpreter of it:

1. **provider id** — the only part of a place that survives the user editing it (#1550). One
   strategy per id rather than one for all of them, so an earlier provider still wins when
   two ids point at different rows.
2. **name** — exact after trim + lowercase.
3. **coordinates** — only when there is no name.

Adds `PlacesService.findMatchingPlaceId` as the public door for an importer that has to link
to the existing place rather than merely skip the candidate. `findDuplicatePlace` stays
private behind it: it also returns `google_ftid` for the bulk importer's backfill, which is
that caller's business and not part of "which place is this?".

`COORD_DEDUP_TOLERANCE` and `externalIdsOf` move to shared and are re-exported from
`places.helpers`, so `collections.service` and `places.service` keep their existing imports
and there is one definition rather than two.

**No behaviour change for either existing caller.** The only semantic difference is on the
branch nothing currently reaches.

Addresses discussion #2129 — this is the first slice of the booking-ingest work designed
there, and the shape (`findMatchingPlaceId` in front, `findDuplicatePlace` private, rule in
shared) is the one agreed in that thread.

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

Marked as a bug fix rather than a refactor: two implementations of one rule disagreeing is a
defect, even though it currently has no reachable symptom. Happy to recategorise if you would
rather it read as preparatory work.

## Checklist
- [ ] I have updated documentation if needed *(no user-facing behaviour to document)*

### Test plan

- `shared/src/place/place-match.spec.ts` — 11 new cases over the rule itself. The one that
  matters is "never offers coordinates for a NAMED candidate, even when it has them".
- `PLACES-SVC-008..013` in `places.service.test.ts` — 6 new cases over `findMatchingPlaceId`
  against real SQL. `PLACES-SVC-010` is the divergence: a named candidate must not match a
  different place at the same coordinates. It fails without this change.
- `npx vitest run tests/unit/nest/places.helpers.test.ts tests/unit/nest/places.service.test.ts tests/unit/nest/collections.service.test.ts`
  — 186 passed, covering both consumers of the moved helpers.
- `npm run typecheck` and `npm run typecheck:tests` — clean.
- `eslint` on the changed files — 0 errors, 0 warnings, no new `any`.
- Coverage: `src/nest/places/**` measures 92.8 / 83.7 / 97.1 / 95.9 against its ratchet of
  91 / 82 / 96 / 94, so the entry is unchanged and not lowered.

### DIFF
--- a/server/src/nest/places/places.helpers.ts
+++ b/server/src/nest/places/places.helpers.ts
@@ -135,9 +135,13 @@ export interface DedupSet {
  *
  * The in-memory half of the matching rule; `PlacesService.findMatchingPlaceId` is
  * the SQL half. Both walk the same strategy list from @trek/shared, so neither can
- * drift into recognising a place the other would not — which is exactly what had
- * happened: the SQL copy reached for coordinates on a named candidate, and this
+ * reach for a KIND of match the other would not — which is exactly what had
+ * happened: the SQL copy fell back to coordinates on a named candidate, and this
  * one deliberately never does (see place-match.ts for why).
+ *
+ * Shared order, not shared comparison: SQLite `lower()` is ASCII-only where
+ * JavaScript's is not, and only unnamed rows contribute coordinates to a dedup
+ * set. `findDuplicatePlace` spells out where the two still answer differently.
  */
 export function isPlaceDuplicate(candidate: PlaceMatchCandidate, dedup: DedupSet): boolean {
   for (const strategy of placeMatchStrategies(candidate)) {
--- a/server/src/nest/places/places.helpers.ts
+++ b/server/src/nest/places/places.helpers.ts
@@ -1,5 +1,11 @@
 import { XMLParser } from 'fast-xml-parser';
 import unzipper from 'unzipper';
+import {
+  externalIdsOf,
+  normalizePlaceName,
+  placeMatchStrategies,
+  type PlaceMatchCandidate,
+} from '@trek/shared';
 import type { Place } from '../../types';
 import type { PlaceWithTags } from '../database/database.service';
 import type { KmlImportSummary } from './kml-import.helpers';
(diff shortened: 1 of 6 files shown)


mauriceboe:
Merged, with a fix pushed onto the branch first.

The behaviour-change claim does not quite hold, and the reason is worth writing down. `isPlaceDuplicate` was handed a mapped candidate but `findDuplicatePlace` got the raw parser item, which spells the provider id `googleFtid` where the rule reads `google_ftid`. Its id strategy was therefore always empty, the name decided the match, and a namesake could take the ftid belonging to the renamed row. Both calls now share one candidate.

That also makes the coordinate branch reachable, contrary to the description. SQLite `lower()` is ASCII-only where JavaScript's is not, so a row stored as `CAFÉ CENTRAL` misses the candidate `Café Central` in SQL while matching it in memory. Before this change the coordinate fallback quietly covered that, sometimes with the wrong row. The cost today is a ftid backfill that does not happen, which I prefer to a backfill landing on a different place, so I kept your order and documented the boundary instead.

Same for `buildDedupSet` collecting coordinates only for unnamed rows while the query considers every row: that is the answer `findMatchingPlaceId` wants for the booking case, so it is stated rather than removed. PLACES-SVC-014 to -017 and PLACE-SVC-028e pin all of it, including the tolerance as a number rather than as a reference to itself.

One for you, not in this PR: `findDuplicateCollectionPlace` in `collections.service.ts` is a third hand-written copy of the same order, and it still falls back to coordinates for a named candidate. That divergence is live rather than latent, so the restaurant and the bar in one building do merge there today. Worth a follow-up now that the strategies exist.

PR #2145, opened 2026-08-29
fix(atlas): wrap long tooltip text instead of overflowing viewport

### LINKED ISSUE #2153: [BUG] Atlas: bucket-list marker tooltip overflows the viewport with long notes
### Pre-flight checklist

### TREK version
4.1.1

### Describe the bug
On the Atlas map, hovering a bucket-list (planned location) marker that has long notes renders the tooltip as a single `white-space: nowrap` line with no width cap. For notes longer than a sentence or two, the tooltip runs off both edges of the viewport instead of wrapping, making the content unreadable.

### Steps to reproduce
1. Go to Atlas
2. Add a bucket-list location with a long `notes` value (a paragraph or more)
3. Hover the star marker for that location on the map
4. See the tooltip text overflow off both edges of the screen

### Expected behavior
The tooltip should wrap and stay within the viewport, capping its height with a scrollbar for very long notes, and staying positioned on-screen even for markers near a screen edge.

### Deployment method
Docker Compose

### Accessing TREK from
Desktop browser

### Browser (if applicable)
Safari 18 (also reproduces in Chromium-based browsers)

### Screenshots
(see linked PR for before/after screenshots)

### Additional context
Fix: liketrek/TREK#2145

BeInLife:
## Summary
- Bucket-list marker tooltips on the Atlas map rendered as a single nowrap line with no width cap, so long notes overflowed off both edges of the viewport.
- Wrap and cap the tooltip's size, with a scrollbar shown only when content actually overflows it, scoped to this tooltip so short region/country tooltips are unaffected.
- The tooltip is now interactive with a delayed close (so the pointer can reach it to scroll instead of Leaflet closing it the instant it leaves the marker's tiny hit area), and its direction/offset is computed from the marker's on-screen position before opening so it stays clear of screen edges.

## Test plan
- [x] Added unit tests in `atlasModel.test.ts` for the extracted pure placement/scroll-need logic (`bucketTooltipWidth`, `bucketTooltipPlacement`, `bucketTooltipNeedsScroll`): centered/no-nudge, left-edge nudge, right-edge nudge, top-edge flip, and the scroll-threshold rounding case.
- [x] Added a bucket-list location with a long `notes` value and hovered the marker on Atlas — tooltip wraps, caps its width/height, and stays within the viewport.
- [x] Verified the scrollbar only appears when the note actually exceeds the height cap (short notes show no scrollbar).
- [x] Verified the tooltip can be scrolled by moving the pointer from the marker onto the tooltip.
- [x] Verified a marker near the top/left/right edge of the map repositions the tooltip to stay on-screen.
- [x] Verified short region/country tooltips elsewhere on Atlas are unaffected.

Closes #2153

### DIFF
--- a/client/src/index.css
+++ b/client/src/index.css
@@ -122,8 +122,29 @@ body { height: 100%; overflow: auto; overscroll-behavior: none; -webkit-overflow
   font-family: inherit !important;
   box-shadow: 0 8px 32px rgba(0,0,0,0.25) !important;
   transition: none !important;
+  max-width: min(280px, calc(100vw - 32px)) !important;
+  white-space: normal !important;
+  word-wrap: break-word !important;
 }
 .atlas-tooltip::before { border-top-color: rgba(10, 10, 20, 0.6) !important; }
+/* Bucket-list marker tooltip only — arbitrary-length user notes need a scroll
+   cap. The scroll happens on .atlas-tooltip-scroll-inner, not this element,
+   so the outer border-radius clips the scrollbar instead of it poking
+   through the rounded corner. overflow-y on the inner div is toggled
+   per-instance in useAtlas.ts, not forced here. */
+.atlas-tooltip-scrollable {
+  width: min(480px, calc(100vw - 32px)) !important;
+  max-width: min(480px, calc(100vw - 32px)) !important;
+  overflow: hidden !important;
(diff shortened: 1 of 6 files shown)


BeInLife:
before:
image
after:
image

Issue #1939, opened 2026-08-14
[BUG] Google Maps API Key Issue


rumexcrisp:
### Pre-flight checklist

- [x] I have searched [existing issues](https://github.com/liketrek/TREK/issues) and this bug has not been reported yet
- [x] I am running the latest available version of TREK
- [x] I have read the [Troubleshooting guide](https://github.com/liketrek/TREK/wiki/Troubleshooting) and my issue is not covered there

### TREK version

v3.4.1

### Describe the bug

I'm using a google maps api key with no access restrictions set (like ip-address or website) and which is allowed to query the Places API, Places API (new), Routing API, etc..

The admin (me) can now search for places in a trip, a user gets a toast notification: `The caller does not have permission`.

### Steps to reproduce

1. Go to admin settings, and fill google maps api key.
2. Click save
3. Search for place as admin -> works
4. Search for place as user -> `The caller does not have permission`

### Expected behavior

Searching a place as a user should work.

### Deployment method

Docker Compose

### Host OS

Unraid 7.3.0

### Accessing TREK from

Desktop browser

### Browser (if applicable)

Chrome 151.0.x

### Relevant logs or error output

```shell
[Google API] #7 searchText → https://places.googleapis.com/v1/places:searchText
Maps search error: Error: The caller does not have permission
    at searchPlaces (/app/server/dist/services/mapsService.js:633:21)
    at process.processTicksAndRejections (node:internal/process/task_queues:104:5)
    at async MapsController.search (/app/server/dist/nest/maps/maps.controller.js:51:20) {
  status: 403}

[Google API] #8 autocomplete → https://places.googleapis.com/v1/places:autocomplete
Maps autocomplete error: Error: The caller does not have permission
    at autocompletePlaces (/app/server/dist/services/mapsService.js:680:21)
    at process.processTicksAndRejections (node:internal/process/task_queues:104:5)
    at async MapsController.autocomplete (/app/server/dist/nest/maps/maps.controller.js:92:20) {
  status: 403}
```

### Screenshots

![Image](https://github.com/user-attachments/assets/273deee4-d356-4a6c-bcf0-89d677eda1cf)

### Additional context

After setting a google maps api key in the admin panel - and clicking the save button - there is no log printed like when updating other admin settings, e.g. `<date> <user-email> admin.permissions_update ip=...`.

I would assume a log line like: `<date> <user-email> admin.api_key ip=...`.

mauriceboe:
[...]

Issue #1611, opened 2026-07-19
[BUG] Synology Photos: "Failed to fetch asset" on both thumbnail and original via unified /api/photos/:id endpoint


Freddiiie:
### Pre-flight checklist

- [x] I have searched [existing issues](https://github.com/liketrek/TREK/issues) and this bug has not been reported yet
- [x] I am running the latest available version of TREK
- [x] I have read the [Troubleshooting guide](https://github.com/liketrek/TREK/wiki/Troubleshooting) and my issue is not covered there

### TREK version

3.4.0

### Describe the bug

After successfully configuring the Synology Photos integration (connection test passes, login works), imported photos in the Journey section fail to display — only broken image placeholders are shown.

### Steps to reproduce

1. Configure Synology Photos under Settings → Integrations → Photo Providers. Test Connection succeeds.
2. Add/import a photo from Synology Photos into a Journey entry.
3. The photo grid shows broken image icons instead of thumbnails.

### Expected behavior

Thumbnail and original image should load via the unified photo endpoint.

Actual behavior:
- Browser shows broken image icon (broken <img> load).
- Network tab shows GET /api/photos/:id/thumbnail → 500, response body {"error":"Failed to fetch asset"}.
- Same result for GET /api/photos/:id/original.
- Server logs (even with LOG_LEVEL=debug) show only the access-log line, no stack trace or underlying error — the actual exception doesn't appear to be logged anywhere, including /app/data/logs/trek.log.

### Deployment method

Docker Compose

### Host OS

Synology DSM 7.3

### Accessing TREK from

Desktop browser

### Browser (if applicable)

Chrome,  Version 150.0.7871.128

### Relevant logs or error output

```shell

```

### Screenshots

<img width="937" height="407" alt="Image" src="https://github.com/user-attachments/assets/a20593bd-576a-4864-b94d-625b714c1896" />

### Additional context

- Docker (docker-compose), TREK behind reverse proxy at custom domain
- Synology Photos on local NAS, accessed via internal IP/hostname, self-signed certificate
- ALLOW_INTERNAL_NETWORK=true set and confirmed working (connection test succeeds)
- "Skip SSL verification" enabled in Synology Photos provider settings
- MFA/2FA enabled on the Synology account
- Tested with photos from both Personal Space and Shared Space — same result for both
- Tested with two Synolgy NAS
- The Synology connection test itself succeeds and returns "Connected."
[...]

PR #2224, opened 2026-09-03
feat(transit): let an admin route transit search through Google

subdee:
## Description
Transitous has no GTFS coverage for much of Asia — a search between two Osaka
subway stops returns "No routes found" because the stops are not in the feed at
all. This adds Google as an optional transit backend behind an admin setting.

Admins pick the backend in **Admin → Settings → Transit Provider**. Google reuses
the Places key the install already has (`maps_api_key`: operator env → instance →
the caller's own row), so there is nothing new to configure. Transitous stays the
default and the fallback: an install that never opens the switch, or that selects
Google before pasting a key, keeps running on Transitous.

**Routes API, not the legacy Directions API** that the discussion named. Every
other Google call in this repo already speaks the v1/`X-Goog-Api-Key` family, and
Google closed the legacy endpoint to Cloud projects created after March 2025 — a
new self-hoster cannot enable it.

Layers:
- **Contract** — `shared/src/admin/admin.schema.ts`: `TRANSIT_PROVIDERS`, the PUT
  body schema, and `ApiKeySource`, which the server's `instance-api-keys.ts` now
  aliases rather than restating (it goes on the wire, so two copies would fork it).
- **Setting** — `server/src/nest/transit/transit-provider.ts`: one `app_settings`
  row, read by `TransitService` per request, written by `AddonsService`. Reads
  fail-safe — a missing row or an unrecognised value resolves to Transitous
  rather than silently billing a key. No migration needed.
- **Backend** — `server/src/nest/transit/google-transit.provider.ts`: Routes API
  for `plan`, Places Text Search for `geocode`, mapped onto the existing compact
  itinerary shape.
- **Dispatch** — `transit.service.ts`: after validation, so both backends are held
  to the same coordinate/mode/transfer contract and return the same 400s.
- **Admin** — `GET`/`PUT /api/admin/transit-provider`, audited.
- **UI + i18n** — a select and two warnings in `AdminSettingsTab.tsx`; seven keys
  across all 23 locales.

Three things worth a reviewer's attention:

**Walking is reshaped.** Google returns walking as turn-by-turn navigation steps
where MOTIS returns one leg per walking segment, so consecutive walk steps are
coalesced into one leg. Without it a two-minute walk arrived as six legs, and a
journey with sliced walks at both ends exceeded `transitItinerarySchema`'s 20-leg
cap and was dropped whole at the MCP boundary. Merging the geometry needs an
encoder (step polylines are delta-encoded from their own first point and cannot
be concatenated), so `transit.helpers.ts` gains a polyline codec.

**Responses name the backend that answered.** The fallback is silent by design,
which left an empty result ambiguous between "this provider has no data here" and
"the provider you picked never ran". The panel's empty state now reads "No
connections found via Transitous." `{provider}` is a placeholder in every locale,
never a baked-in name — hardcoding it per locale is the #1611 bug class, and
`i18n-placeholders.spec.ts` enforces it.

**The admin panel warns about keys.** Selecting Google with no key that resolves
changes nothing, so the panel says so. It also warns when only the admin's *own*
key is set: the resolver's last step is the caller's own row, so a personal key
serves that admin while every other member silently gets Transitous — the #1939
shape one layer up.

Cost shaped the implementation, since Google bills where Transitous does not. Both
field masks are cut to the cheapest SKU tier that still answers the question — the
station picker deliberately does *not* reuse `SEARCH_TEXT_FIELD_MASK`, whose
rating/website/phone fields bill Text Search at Enterprise. Cache TTLs are longer
than the Transitous path's 60s, and route alternatives ride along inside the one
request that was already billed.

Addresses discussion #1699

## Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update

## Checklist


mauriceboe:
I think you need to let Claude know not to include any more line breaks in PRs. 😅

mauriceboe (approved):
LGTM

PR #2237, opened 2026-09-04
feat(maps): add a Google Maps renderer as a fourth map provider

MElkmeshi:
## Description

Adds `google-maps` as a fourth map provider alongside Leaflet, Mapbox GL and MapLibre GL, selectable in **Settings → Map** with its own API key field and an instance-wide admin default.

Google's API is a different shape from the two GL engines. Those share one API surface, which is why they share one component with the engine injected as a `gl` prop — Google has no style spec, no `addSource`/`addLayer`, and its own marker and camera model, so it cannot become a third engine behind `MapViewGL`. It gets its own component and reuses the parts that make a TREK map look like a TREK map: the pin element, the popup markup, the viewport maths and the photo cache are lifted into shared modules (`placeMarkerElement.ts`, `categoryIcon.ts`, `usePlacePhotos.ts`) that `MapViewGL` now also uses, rather than copied.

The SDK is loaded from Google's hosted loader rather than bundled, because that is the only form the renderer is licensed in. That brings a CSP entry for `maps.googleapis.com` and a loader that handles the two ways a key fails: a network error, and a served-but-unauthorised key that Google reports through `gm_authFailure` rather than a failed request.

### Two findings worth the reviewer's time

**`AdvancedMarkerElement` is unusable on a raster map.** It is the obvious modern choice and it is the wrong one here. It repositions in two phases: it rides a coarse transform during a drag, then on idle snaps back to its pre-drag screen position and animates forward again. Measured against an `OverlayView` pin held at the same coordinate on the same map under one real drag:

| t | AdvancedMarkerElement | OverlayView | offset |
|---|---|---|---|
| 56851 | 155,196 | 162,224 | -7,-28 |
| 57050 | 47,82 | 54,110 | -7,-28 |
| 57252 | **155,196** ← back at its pre-drag spot | 18,72 | 137,124 |
| 57401 | 11,44 | 18,72 | -7,-28 |

Vector rendering avoids it, but that needs a Map ID created in the Google Cloud console, which a self-hosted instance cannot assume — and requesting `RenderingType.VECTOR` against `DEMO_MAP_ID` reports VECTOR and then renders an empty map (that attempt and its revert are both in the history). So the pins are drawn by a single `OverlayView` that positions its own DOM from the projection on each draw.

**`normalizeProvider` made the settings tab unsafe for any new provider.** It coerces anything unrecognised to `leaflet`, so an instance running `google-maps` showed Leaflet as selected and the next save silently wrote that over the user's real choice. That is why the Map tab changes are part of this PR rather than a follow-up. The GL-only sections are now gated on an explicit `isGl` check rather than "not leaflet" — those were the same thing until Google arrived.

### What works

Verified against a live key and a 34-place trip: basemap, place pins with their photos and day-order badges, hover card at full resolution, click-to-select, day route polylines, fit-to-places, and the compass. RTL basemap labels are correct here for free, since Google shapes them server-side.

### What is deliberately not in scope

- **Unported surfaces:** reservations/transport lines, POI explore, plugin map layers and markers, marker clustering, 3D/terrain, and the Journey and Atlas maps. The provider falls back cleanly; these simply do not render on Google yet.
- The small **map preview** in Settings still draws Leaflet when Google is selected — that widget only knows Leaflet and GL.

### Two things that may make this unwanted regardless of the code

1. **Offline.** Google's terms prohibit caching or pre-downloading tiles. TREK is offline-first and precaches up to 12,288 tiles per provider, so this can never be an offline provider. That is a licensing limit, not something implementable around.
2. **Cost.** It needs a billing-enabled key and bills per map load, where every current provider is free or has a free tier. `google_maps_api_key` is therefore on the managed-instance locked list, next to the Mapbox and CARTO keys.

I would rather you weighed both before anyone spends more time here.

None — there is no approved feature discussion, and I have not pitched this in `#github-pr`. I know that does not meet the bar; close it if you would rather it started as a Discussion and I will move it there. The RTL fix that came out of the same work went through as #2236 and is merged.

## Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update

## Checklist

## Test plan

- Unit tests for the SDK loader (including the `gm_authFailure` path and a refused second key), the compass adapter, and the pin/hover-card image selection.
- Full server suite green (10,051 tests). Two parity tests correctly caught that adding a key to `ENCRYPTED_SETTING_KEYS` is not self-contained — the encryption-rotation script keeps its own key list, and the managed locked list is pinned verbatim. Both are registered rather than loosened.
- New settings strings translated in all 23 locales; `i18n:parity:strict` passes.
- All CI checks green: server and client coverage, both type/lint jobs, i18n parity, plugin facts.
- Manually verified end to end against a live key on a self-hosted instance: pins, photos, hover, selection, panning, and a Google map rendering Arabic correctly.

### DIFF
--- a/client/package.json
+++ b/client/package.json
@@ -74,6 +74,7 @@
     "@testing-library/react": "^16.3.2",
     "@testing-library/user-event": "^14.6.1",
     "@trivago/prettier-plugin-sort-imports": "^6.0.2",
+    "@types/google.maps": "^3.66.2",
     "@types/leaflet": "^1.9.8",
     "@types/node": "^25.9.3",
     "@types/react": "^19.2.15",
--- a/client/src/components/Map/MapViewAuto.tsx
+++ b/client/src/components/Map/MapViewAuto.tsx
@@ -2,7 +2,8 @@ import { Suspense } from 'react'
 import { useSettingsStore } from '../../store/settingsStore'
 import { MapView } from './MapView'
 import ErrorBoundary from '../shared/ErrorBoundary'
-import { MapViewGLMapbox, MapViewGLMaplibre } from './glLazy'
+import { MapViewGLMapbox, MapViewGLMaplibre, MapViewGoogleLazy } from './glLazy'
+import { basemapLanguage } from './glProviders'
 
 // Auto-selects the map renderer based on user settings. Keeps the existing
 // Leaflet MapView untouched so the Mapbox GL variant can mature iteratively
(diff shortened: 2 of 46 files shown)


mauriceboe:
Closing this, and I want to be straight with you about why rather than sending you off to fix fifteen review points first.

It is the cost side, not the code. Google bills per map load and forbids tile caching, so this can never be an offline provider, while Leaflet, MapLibre and Mapbox are free or have a free tier. And on the Google renderer the map does less than the one people already have: no reservation and transport lines, no POI explore, no plugin layers or markers, no clustering, no satellite, no right-click to add a place. So someone who pays for it ends up with a weaker TREK map, just on a Google basemap. The one real win, Google's server-shaped RTL labels, already landed separately in #2236.

Against that stands a permanent cost for a self-hosted project: a fourth branch in every provider switch, a Google script origin in every instance's CSP, and a billing relationship. That is not a trade I want to lock in.

What I would take right away is the part that has nothing to do with Google: `placeMarkerElement.ts`, `categoryIcon.ts` and `usePlacePhotos.ts` pull 147 lines out of `MapViewGL.tsx` without changing behaviour, and the `isGl` gate in `MapSettingsTab.tsx` fixes a latent bug we have regardless. If you open that as its own small PR, I will merge it.

Sorry it took a full renderer to get to this conversation. That is on us for not having the provider question written down anywhere.

PR #2268, opened 2026-09-06
feat(budget): attach receipts and invoices to expenses

dersimoezdag:
## Description
Allows attaching receipts and invoices (images & PDFs) directly to expenses in both Desktop and Mobile views.
- Users can upload multiple receipts or invoices per expense via drag-and-drop or file picker.
- Displays an receipt chip (`[ 🧾 Beleg ]` / `[ 🧾 Belege (3) ]`) in the expense ledger rows.
- Clicking the chip opens an in-app viewer supporting both images and PDFs, complete with tab switching and keyboard navigation (← / →) for multi-receipt expenses.
- Automatically soft-deletes orphan receipts to the trip trash when an expense is deleted or an attachment is unlinked, preventing orphaned receipt files.

none

## Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update

## Checklist


mauriceboe:
Thanks for this, and for building it the right way round. You hung it on the existing `trip_files` / `file_links` infrastructure instead of opening a second upload path, so the storage driver, the size and extension limits, the trash and the signed downloads all come along for free. Desktop and mobile both get it, the viewer with tabs and arrow keys is a nice touch, and cleaning up orphaned receipts when the expense goes is the part people usually forget. The direction is right and I want this in.

A few things need to happen before it can go in, and none of the CI jobs that would have told you have run yet (only `check-target` did, so lint, typecheck, tests and i18n parity never fired on this branch).

**Blocking**

1. **The client typecheck fails.** `BudgetItemReceipt` is used in `MCostsTab.tsx:82` and `:557` but never imported, so `cd client && npm run typecheck` dies with TS2304 and the branch does not build.

2. **`budget_item_id` never reaches the upload handler.** You extended `FilesService.createFile` / `updateFile` / `findForeignLinkTarget`, but `files.controller.ts:151` and `:164` still pass only `reservation_id` and `place_id`. The client's `fd.append('budget_item_id', ...)` is silently dropped: the file uploads but is neither written to `trip_files.budget_item_id` nor linked. It looks like it works in the edit flow only because saving re-links through `receipt_file_ids` afterwards. The new cross-trip check has the same gap: it runs on the link route but not on upload or update.

3. **i18n parity will fail.** Nine new keys landed in `en` and `de` only; the other 21 locales need them, and `i18n:parity:strict` is a hard gate. Three of the nine (`costs.receipt`, `costs.receiptCount`, `costs.uploadReceipt`) look unused, worth dropping before translating them 23 times.

**Worth fixing while you are in there**

4. **Orphaned uploads.** On create, files upload before the expense exists, so a failed `addBudgetItem` leaves them unlinked and no cleanup path reaches them (`trashOrphanReceipt` only runs from an existing item). In the edit modal it is the mirror image: the upload happens on selection, and Cancel leaves the file behind.

5. **Coverage.** About 190 new server lines with one test. `budget` is pinned at 95 lines and `files` at 98 in `server/vitest.config.ts`, so this will drop below. `trashOrphanReceipt`'s four bail-out branches, the update sync and the delete path are the ones that matter.

6. **Index drift.** `schema.ts` creates `trip_files.budget_item_id` on fresh databases, but the migration creates `idx_trip_files_budget_item_id` *inside* the `pragma_table_info` guard. A fresh install gets the column and never the index; an upgrade gets both.

7. **The relationship is modelled twice**, once as `trip_files.budget_item_id` and once as a `file_links` row, which is why the reads need a COALESCE plus DISTINCT plus a manual dedupe and why five write sites have to stay in step by hand. Not a blocker, but if you want to simplify, `file_links` alone would carry it.

Happy to look again once CI is green on it.

dersimoezdag:
Thanks for the nice review and helpful feedback! I've addressed all issues and your suggestions in the latest push:

### 1. Client Typecheck (`BudgetItemReceipt`)
- Added the missing import `BudgetItemReceipt` in `MCostsTab.tsx`. The client code now builds and typechecks cleanly with zero errors in feature files.

### 2. Files Controller Wiring & Cross-Trip Checks
- Passed `budget_item_id` through to `createFile` and `updateFile` in `FilesController`.
- Wired `assertLinkTargets` across `upload`, `update`, and `link` endpoints to prevent cross-trip IDOR when referencing budget items.

### 3. i18n Strict Parity
- I removed the 3 unused keys (`costs.receipt`, `costs.receiptCount`, `costs.uploadReceipt`).
- Synchronized the 6 active receipt keys across all 23 locale files (`shared/src/i18n/*/budget.ts`).
- `npm run i18n:parity:strict` passes cleanly.
- Translated the keys with help of ai.

### 4. Staged Uploads / No More Orphaned Files
- Implemented a staging pattern (`pendingReceiptFiles`) on both Desktop (`CostsPanel.tsx`) and Mobile (`MCostSheet.tsx`).
- Selected files are kept local in state and are only uploaded when the user actually hits "Save".
- Clicking "Cancel" cleanly discards the selection without touching the server/storage.
- If an error occurs during the submit pipeline after upload, any newly uploaded files are permanently deleted as a rollback.

### 5. Dual-Modeling & Index Drift Cleaned Up
- Dropped the duplicate `budget_item_id` column from `trip_files` in `schema.ts` and `types.ts`.
- Receipt linking is now exclusively modeled via `file_links.budget_item_id`, eliminating duplicate write sites and queries needing `COALESCE`/`DISTINCT`.
- Updated migration 206 to only add `budget_item_id` and the corresponding index to `file_links`.

### 6. Test Coverage
- Added full unit and DB test suites in `budget.service.db.test.ts` and `files.controller.test.ts`:
- Covered all 4 bail-out branches in `trashOrphanReceipt` (non-existent, legacy link, external link, and orphan cleanup).
- Tested receipt linking on create, sync on update, and cleanup on expense deletion.  
- Added positive & cross-trip negative assertion tests for the files controller.

The branch has been updated with these fixes. Ready for another look whenever you have time!

PR #2280, opened 2026-09-07
Fix/immich picker ux

xvilo:
## Description
The photo/video picker for adding Immich media to a journey location has been rough to actually use:

- Photos weren't reliably in chronological order — Immich's search response order isn't guaranteed, so the grid could feel shuffled.
- Scrolling down to trigger lazy-load kept you at the bottom, so newly loaded photos landed above your current view instead of where you'd expect — it looked like nothing had happened.
- Nothing distinguished a video thumbnail from a photo, so you couldn't tell what you were selecting at a glance.
- Clicking "Add" gave no feedback — no spinner, no disabled state while the request was in flight — so a slow save looked like a dead click (and nothing stopped a double-submit).

This branch tackles the easy, high-value wins without touching video preview (out of scope for now):

1. Photos now default to newest-first when there's no location context, and each loaded page is sorted once and appended rather than the whole list being re-sorted on every load — so new photos always land at the bottom instead of reshuffling what's already on screen.
2. A small play-icon badge on the thumbnail now marks videos at a glance.
3. The Add button shows a spinner and disables itself while the add is in flight, so it's clear the click registered.

Four small, scoped commits, tests/typecheck/lint all green

I made a small message on Discord

## Type of Change
- [x] Bug fix

## Checklist

### DIFF
--- a/client/src/components/Journey/JourneyDetailPageProviderPicker.test.tsx
+++ b/client/src/components/Journey/JourneyDetailPageProviderPicker.test.tsx
@@ -1,4 +1,4 @@
-// FE-JRN-PICKER-001 to FE-JRN-PICKER-020
+// FE-JRN-PICKER-001 to FE-JRN-PICKER-021
 
 import { describe, it, expect, beforeEach, vi } from 'vitest'
 import { http, HttpResponse, delay } from 'msw'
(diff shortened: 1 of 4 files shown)


mauriceboe:
Thanks for this, and sorry for the slow reply. Three of the four commits are things the picker genuinely needed, so rather than send it back I pushed the last mile onto your branch.

The double submit was a real defect: `onAdd(...)` was called without `await` even though the prop was already typed as `=> Promise<void>`, so two quick clicks fired two complete upload rounds. Good catch. The `takenAt` fallback is careful work too, the comparator is total, transitive and deterministic and it does not mutate the incoming array. And `mediaType` is the right source for the badge, it already flowed through to the selection before this PR.

What I changed on top:

**The sorting moved back into a memo.** Doing it inside `searchPhotos` and `loadAlbumPhotos` looks equivalent and is not. `contextLocation` is live editor state, written by the place search and by "use my location" while the picker stays mounted, and the picker key is only provider plus date, so nothing remounts it. Open an entry with no location, let the photos load, then search for Rome: the caption flips to nearest-first and the grid does not move. The appending search made it worse, since it sorted only the page it had just fetched, so from page two on the grid was a run of sorted blocks and a photo from page two could land under a day heading page one had already closed. `FE-JRN-PICKER-021` pins it and is red without the memo.

**The video badge got `z-10` and `pointer-events-none`.** It shares a stacking context with the city gradient right after it, so on the geotagged tiles the gradient was winning. Nothing was hidden, the gradient is transparent, but the play glyph came out duller than it should. `pointer-events-none` is what the eight existing video overlays in the journey code all set.

Also rebased onto `dev`, which the branch was a good way behind.

One note for next time rather than a complaint: the prettier commit is 483 of the 529 lines. I checked it and it is genuinely logic free, but it is unusual for this repo and it makes `git blame` on those two files harder. Happy to take the functional part on its own in future.

Merging once CI is green.

PR #2319, opened 2026-09-10
feat(maps): draw every travel day's route at once, with the trip total

subdee:
## Description
Adds a whole-trip route overview to the planner map: one toggle draws every travel day's
route at once, each day in its own colour, with a card listing each day's stage — the
modes it is travelled in and its distance — under the trip's total.

The day route was only ever built for the selected day, inside a callback in
`useRouteCalculation`. That builder is lifted unchanged into a React-free
`client/src/components/Map/dayRoutePlan.ts:35`, and
`client/src/components/Map/useTripRouteOverview.ts:77` runs it over every day — so the
overview draws exactly what the day view draws, rather than a second copy of the same
rules. Legs go out a few at a time and share RouteCalculator's cache; one that the router
refuses keeps its straight line and adds no distance. Both shells use one pill and one
card (`TripRouteOverview.tsx`), and the renderers take a parallel `routeColors`
(`MapView.tsx:1039`, `MapViewGL.tsx:1832`).

Nothing is persisted and no API surface is added — this derives entirely from days,
assignments and reservations already in the store. Road trip mode already draws the whole
trip its own way, so the overview stands down there. Also folds the floating map controls
onto one drop shadow (`mapControlShadow.ts`): the frosted ones were using a panel shadow
that all but disappears over map tiles. `wiki/Map-Features.md` documents the new control.

Addresses discussion #1736 (also folds in #1726)

## Type of Change
- [x] New feature

## Checklist
- [x] I have added/updated tests


mauriceboe (dismissed):
lgtm @subdee

mauriceboe (approved):

Issue #2066, opened 2026-08-27
[BUG] PDF export doe not respect 12-hour time format setting


claymcgovern:
### Pre-flight checklist

- [x] I have searched [existing issues](https://github.com/liketrek/TREK/issues) and this bug has not been reported yet
- [x] I am running the latest available version of TREK
- [x] I have read the [Troubleshooting guide](https://github.com/liketrek/TREK/wiki/Troubleshooting) and my issue is not covered there

### TREK version

4.0.0

### Describe the bug

I have my individual user preference set to 12 hour time format and the default set in the Admin section set to 12 hour time format, but when I export to PDF the times are all in 24 hour format.

### Steps to reproduce

1. Set both individual user and admin default time format to 12 hour format
2. Export PDF

### Expected behavior

12 hour format in PDF

### Deployment method

Docker Compose

### Host OS

Ubuntu 24.04

### Accessing TREK from

Desktop browser

### Relevant logs or error output

```shell

```

### Screenshots

<img width="787" height="420" alt="Image" src="https://github.com/user-attachments/assets/e470c472-bb03-4004-b4a9-c05c62110453" />
<img width="1214" height="726" alt="Image" src="https://github.com/user-attachments/assets/34c96098-6905-443c-9165-dc7b96937539" />

mauriceboe:
Confirmed and fixed. The PDF is built as an HTML string outside React, so no time format ever reached it and it printed the stored column at all four clock sites — and for a place time or a check-in stored with a meridiem it went wrong in the other direction too, showing "3:00 PM" to a 24h reader. Both export buttons now hand the setting over the way they already hand over the locale. Fixed in #2074, shipping in the next release.

PR #2330, opened 2026-09-11
Feat/pdf trip route map

subdee:
## Description
Puts the trip's route map into the exported PDF, completing discussion #1736 — the part
left out of #2319.

After the cover comes one map of the whole trip: every planned day in its own colour with
its stops marked, a legend naming each day and the distance it covers, a scale bar, and the
trip's total distance both beside the map and as a new **Distance** tile on the cover.

The backdrop is the instance's own basemap, rendered offscreen through MapLibre and baked
into the document as a PNG (`client/src/components/PDF/tripMapImage.ts`). It has to be a
picture rather than a live map: the preview runs in a sandboxed iframe with no
`allow-scripts`, and printing starts on the click, so nothing may still be loading by then.
Routes are projected with the map's own `project()` so the lines sit on the roads rather
than drifting at the frame edges. OpenFreeMap is the default style and needs no token.

When that cannot be drawn — no WebGL, no connection, a style that will not load — it falls
back to `tripMapSvg.ts`, which draws the bundled country outlines from `countryShapes.ts`
with no network and no key. Both grounds share one `buildRouteOverlay()` so the route on
top looks identical either way.

Routing reuses the day builder rather than a second copy of it:
`useTripRouteOverview`'s core moved into `client/src/components/Map/tripRouteGeometry.ts`,
leaving the hook a thin React wrapper and letting the export call `routeTrip()` directly.
It is time-boxed at 8s — legs that answer become real roads, the rest stay straight lines,
so a long trip with a cold cache cannot hold an export hostage. A failure never costs the
reader their itinerary: the map is dropped and the reason logged under `[tripPdfMap]`.

`distanceUnit` is threaded like `timeFormat` (prop plus store fallback, #2066) and passed
from both export call sites.

Addresses discussion #1736 (also folds in #1726)

> **Depends on #2319** — built on `feat/trip-route-overview`, which is still open. Until
> that merges this PR's diff includes its commit. Both branches are behind `dev`
> (currently `a8e4c455`) and want rebasing before merge.

## Type of Change
- [x] New feature

## Checklist
- [ ] My branch is up to date with `dev`
- [x] I have added/updated tests


mauriceboe (approved):

mauriceboe:
@subdee lgtm

PR #2342, opened 2026-09-12
feat(admin): an instance-wide switch for whether the model reads images

Rem7474:
## Description

`multimodal` has been on the AI Parsing addon config for a while: stored, passed straight back on save, **settable by nobody**. Each user has the toggle in their own settings, but the instance config — which wins over them — did not. So on any instance where an admin configured the model, nothing could say it reads images.

**The switch**, under the model field in both admin screens, for every provider:

- **Local provider — TREK asks, it does not guess.** Ollama's `/api/show` reports what a model can do, and that answer sets the switch and says so. Verified against Ollama 0.32.15:
  ```
  qwen3.5:4b → ['completion', 'vision', 'tools', 'thinking']
  qwen3:8b   → ['completion', 'tools', 'thinking']          ← no vision
  ```
- **Cloud provider, or a local server that cannot answer** — a guess from the model family (`qwen3.5`, `llava`, `minicpm-v`, `-vl`, GPT-4o/5, Claude, Gemini), erring toward letting a model try.
- **It stays an override.** Going against what the server reports, or turning it on for an unknown model, warns — it never blocks Save.

**Server side**, the same answer in one place: `LlmParseService.readsPhotos()` (switch, then the local server, then the id), `GET /api/llm/capabilities` for the signed-in caller — which the public `/api/health/features` cannot give, since it has no idea whose model is configured — and the admin-only `GET /api/admin/llm/local/capabilities` the screens use, forbidden in managed mode like its `models`/`pull` siblings.

The two admin shells render their own markup over one hook (`useModelVision`), so the logic is written and tested once.

**Deliberately not here:** nothing that reads a document as an image. Extraction still sends every provider but Anthropic the extracted text, exactly as #2331 now documents — this PR makes the capability known and askable, so a feature can stop offering a photo that would only come back refused. Nothing consumes `/api/llm/capabilities` yet.

Addresses discussion https://discord.com/channels/1488298068427411591/1542235727033540728

## Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [x] Documentation update

## How to test

1. **Admin → Addons → AI Parsing**, provider **Local**, model `qwen3.5:4b`: the switch comes up **on** and the panel says *The server reports that qwen3.5:4b reads images.*
2. Select `qwen3:8b`: **off**, and the panel says it does not. Turn it on → a warning that this goes against the server; **Save still works**.
3. Provider **OpenAI**, model `mistral-large`: **off**, no request to the local server. Turn it on → *mistral-large is not known to read images…*
4. Local provider with a dead base URL: no note, no error toast, the id guess applies.
5. Save, reload: the state persists (it rides in the addon config as `multimodal`).
6. `curl -b <session> localhost:3001/api/llm/capabilities` → `{"configured":true,"photos":true}` for a vision model, `photos:false` for a text-only one; unauthenticated → 401.
7. Same in the mobile admin screen (an `MToggle`, `role="switch"`).

```bash
npm run build --workspace=shared
npm run test --workspace=shared -- --run src/admin/admin.schema.spec.ts
cd server && npx vitest run tests/unit/nest/llm-parse tests/unit/nest/validate-managed-routes.test.ts tests/unit/nest/route-inventory.test.ts
cd client && npx vitest run src/components/Admin/useModelVision.test.ts src/components/Admin/AddonManager.test.tsx tests/unit/mobile/admin/MAdminAddonManager.test.tsx
```

## Checklist
- [ ] My branch is [up to date with `dev`](https://github.com/liketrek/TREK/wiki/Development-environment#3-keep-your-fork-up-to-date) *(stacked on #2331; rebased onto `dev` when it lands)*
- [ ] This PR targets the `dev` branch, not `main` *(targets #2331's branch until it lands)*

### DIFF
--- a/client/src/api/client.ts
+++ b/client/src/api/client.ts
@@ -613,6 +613,13 @@ export const adminApi = {
   // Local LLM (Ollama) management for the AI-parsing addon.
   llmLocalModels: (baseUrl: string): Promise<{ models: { name: string; size: number }[] }> =>
     apiClient.get('/admin/llm/local/models', { params: { baseUrl } }).then(r => r.data),
+  /**
+   * What the local server says a model can do. `capabilities: null` means the
+   * server could not answer (unreachable, unknown model, or an Ollama too old to
+   * report it) — the caller falls back to guessing from the id.
+   */
+  llmLocalCapabilities: (baseUrl: string, model: string): Promise<{ capabilities: string[] | null }> =>
+    apiClient.get('/admin/llm/local/capabilities', { params: { baseUrl, model } }).then(r => r.data),
   /** Pull a model, streaming Ollama's NDJSON progress to `onProgress`. */
   llmLocalPull: async (
     baseUrl: string,
--- a/client/src/components/Admin/AddonManager.test.tsx
+++ b/client/src/components/Admin/AddonManager.test.tsx
@@ -72,7 +72,9 @@ beforeEach(() => {
   seedStore(useSettingsStore, { settings: { dark_mode: false } });
   vi.spyOn(useAddonStore.getState(), 'loadAddons').mockResolvedValue(undefined);
   server.use(
-    http.get('/api/admin/addons', () => HttpResponse.json({ addons: [] }))
+    http.get('/api/admin/addons', () => HttpResponse.json({ addons: [] })),
+    // What Ollama says about a model: nothing, unless a test says otherwise.
+    http.get('/api/admin/llm/local/capabilities', () => HttpResponse.json({ capabilities: null })),
   );
 });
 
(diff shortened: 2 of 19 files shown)


mauriceboe (commented):
Thanks for this, and for checking `/api/show` against a real Ollama. I'm holding it for now, because the switch doesn't change anything yet and the new endpoint has no caller.

- `multimodal` isn't read anywhere in extraction. `parse()` sends native bytes only for Anthropic PDFs and extracted text for everything else, so the hint ("a vision model is handed the document itself") and both warnings describe behaviour TREK doesn't have.
- The override only works one way. `readsPhotos` treats `multimodal: false` as "not set" and falls through to Ollama or the id guess, and on the client the Ollama answer overwrites the stored value on every mount. An admin can't keep it off for gpt-4o or a local vision model, and a saved choice can come back different after a reload. Reading a stored `false` as "off" doesn't work either, since the old admin screens wrote `false` on every save, so this needs an explicit auto/on/off.
- `GET /api/llm/capabilities` has no caller, but for a local provider every request posts to Ollama. The `/api/show` body is read with an uncapped `res.json()` and an `as` cast (`readCappedJson` in `utils/cappedFetch.ts` is the pattern), `ollamaRoot()` sits outside the try so a bad stored URL becomes a 400, and the lookup fires on every keystroke in the model and URL fields.
- The new strings are English literals, including the sentences built in the hook. The per-user twin already has keys in all 23 locales (`settings.aiParsing.multimodal`), and the desktop switch should be `ToggleSwitch` like the rest of the screen.
- The id regex misses current vision models (gpt-4.1-mini, o4-mini, gemma3, pixtral, llama4, qwen2.5vl), while the recommended-model list already carries a `vision` flag.

I'd rather see the capability land together with the feature that actually sends a photo, so the switch has something to control and the endpoint has its caller. Happy to look at it again in that shape.

PR #2351, opened 2026-09-12
feat(managed): a boot hook for an install that reports to its own collector

isaiastavares:
## Description

TREK sends no telemetry, and this does not change that. What it adds is the seam an operator needs when *they* want their own observability, and the documentation for the part that needs no TREK code at all.

**The hook.** `client/src/managed/index.tsx` could already attach routes, nav entries and admin tabs, but nothing that runs at startup — so an install that wanted error reporting or product analytics had to patch `main.tsx`, a seam to re-apply on every rebase. `onAppBoot()` is that missing piece: called once, before the app renders, empty in this repository and empty in every build of it.

**The gate, and where it lives.** `client/src/managed/boot.ts` reads the visitor's Do Not Track / Global Privacy Control signal and returns before reaching the hook. It is deliberately **not** part of the attachment point: an install replaces `index.tsx`, `boot.ts` stays, so the refusal is a guarantee rather than a request. Reading the signal first also means nothing is half-done — no script loaded, no request sent, then regretted.

**What keeps the promise honest.** `FE-MANAGED-009` watches `fetch`, `XMLHttpRequest.open`, `sendBeacon` and `Element.appendChild` while calling the shipped `onAppBoot()`, and fails if anything ever leaves the tab. The point is not that the function body is empty today; it is that nothing goes out when it runs.

**The wiki page.** `Observability.md` writes down what an operator currently has to discover the hard way:

- OpenTelemetry needs **no TREK code and no `OTEL_*` support in TREK**. Zero-code instrumentation attaches to a Node process from the outside, and the image's entrypoint is a plain `node` invocation, so `NODE_OPTIONS` reaches it and the standard variables work untouched. The real obstacle is packaging — the official image ships without npm — so the page shows the derived-image `Dockerfile`. The same `--require` trick covers an error reporter with its own SDK.
- On the client, `script-src 'self'` blocks a pasted analytics snippet outright and `connect-src` is an explicit allow-list. The page says so, and says the two ways around it that need no CSP change: bundle the SDK, and route ingest through the instance's own origin (PostHog reverse proxy, Sentry `tunnel`).

No new dependencies, server untouched, and an install that configures nothing takes exactly the code path it takes today.

Pitched in `#github-pr` on Discord (opt-in, vendor-neutral observability that stays off without configuration).

The design moved during implementation and it is worth flagging: the pitch floated `OTEL_*` support in the server and a `VITE_ANALYTICS_SCRIPT_URL`. Neither is here. The first turned out to need no TREK code — hence the wiki page instead of a dependency. The second cannot work: a third-party script URL is refused by TREK's own `script-src 'self'`, silently, in the browser. What is left is the one piece that genuinely was missing.

## Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [x] Documentation update

## Checklist

## Test plan

```
npx vitest run src/managed src/utils/doNotTrack.test.ts   # 21 passed
npm run typecheck                                          # clean
npx eslint src/managed src/utils/doNotTrack.ts src/main.tsx # clean
npm run lint:pages                                         # Page pattern OK
```

Coverage on the new code is 100% of statements, branches and functions (`boot.ts`, `doNotTrack.ts`; `onAppBoot` is covered by `FE-MANAGED-009`).

The full client suite is green apart from five pre-existing failures on this machine — `MAdminMcpTokensPanel`, `MAdminUsersSection`, `OfflineTab` and `MSettingsOffline` assert US-locale date strings (`4/1/2025`) and fail on a non-US ICU default. None of them import anything this PR touches.


mauriceboe:
Thanks for the detailed write-up. We're going to pass on this one.

TREK sends no telemetry, and we want that to hold for what the project supports, not only for what the default build does. A boot hook whose documented purpose is wiring analytics and error reporting into the frontend, together with a wiki page on how to do it, would make that a supported path of TREK. Even opt-in and behind Do Not Track, that is not a direction we want to take.

`client/src/managed/` is also internal. It isn't a public extension point, it can change or disappear without notice, and we don't want it documented as something to build on. The same goes for the OpenTelemetry page: the wiki ships inside the image as in-app help, and we only document setups we build and test ourselves.

Closing this. If you run your own build of the client, patching it there is the way to go.

PR #2352, opened 2026-09-12
feat(config): APP_NAME as the one place the product name lives

isaiastavares:
## Description

`"TREK"` was copied as a literal across dozens of call sites on the server. Changing the name — or letting an instance run under its own — meant hunting strings one by one, with the usual chance of missing some. This is that cleanup: **one name, one place.**

`APP_NAME` now sits next to `appVersion` in `deriveApp()`, defaults to `TREK`, and every surface that announces the install to somebody else derives its text from it:

| | where |
|---|---|
| Outbound `User-Agent` | Nominatim, Overpass, Wikipedia, Wikimedia, Atlas, transit; plus the GitHub releases API, the plugin registry, plugin downloads and the wiki fetch |
| Calendars | the `PRODID` of every downloaded `.ics` and every subscribable feed, and the calendar name and filename an untitled trip falls back to |
| GPX | the `creator` attribute of a downloaded file |
| OpenAPI | the title and description at `/api/docs` |
| MCP | the server name (was repeated in four files, now one helper), the `realm` of every `WWW-Authenticate`, the `invalid_target` description, and the `initialize` instructions every AI client reads |
| MCP | …and the description of every tool, resource and prompt |
| Email | the subject prefix, the header block and the `From` display name |
| Notifications | the text a user reads on every channel — email, in-app, push — and the `source`, Discord footer and Slack link label of an outgoing webhook |
| Platform | the name a passkey prompt shows and the issuer an authenticator app shows, for credentials enrolled from here on |
| Messages | errors and notices that name the app: a rejected backup upload, an unreachable relay, the first-run console banner |

`MAIL_FROM_NAME` overrides just the `From` display name, for a relay whose verified sender identity has to differ. It is added only when the configured sender is a bare address — one that already spells a name out (`Acme <no-reply@acme.example>`) is used verbatim. The object form is deliberate so nodemailer does the RFC 5322 quoting; a name with a comma can never break the header.

**With `APP_NAME` unset, every one of these emits byte-identical output to what it emitted before.** That is the property the tests are built around: an upgrade must not quietly change the `User-Agent` an install sends to OpenStreetMap, or the `PRODID` inside calendars people have already subscribed to.

### One transform instead of twenty `readEnv()` calls

MCP descriptions and the external-notification locales are static literals written next to what they describe, and there are a lot of them. Rather than turn each into a template literal, they carry an `{appName}` placeholder:

- `nest-mcp` gains an optional `describe` hook on `McpModuleOptions`, applied to tools, resources, templates and prompts alike at **attach** time. The generic layer stays generic — it knows about a text transform, not about TREK — and the value is read per session rather than frozen when the decorator was evaluated.
- The locales resolve theirs in `getEventText()`, the single entry point `NotificationsService` uses for the email, the in-app row and the push payload, so all three say the same thing.

### Two details the formats forced

- **`PRODID` escaping.** RFC 5545 §3.2 makes it a TEXT value: a name carrying `;`, `,`, `\` or a newline ends the property early and corrupts every following line of the file. The escape closure that already lived inside `buildTripCalendar` is hoisted to module scope and reused, so there is one copy of the rule rather than two.
- **The `User-Agent` keeps `(https://github.com/liketrek/TREK)`.** That link is provenance, not identity — `maps.helpers.ts` explains why these APIs want a UA that identifies the caller (#1309), and the operator they need to reach is identified by the instance URL that gets appended.

### Deliberately left alone

`TREK_*` variables, the `trek_session` cookie, table and column names, `trek-plugin-sdk`, the `trek` field of a plugin manifest, the `X-TREK-Instance` header, and the Immich `deviceId` used to correlate uploads — renaming any of those breaks a deployment or existing rows.

Plugin host-compatibility messages ("requires TREK 3.2.0+") and the new-version notification keep saying TREK, because they are about the version of the software rather than about the instance. So does the TREK Places index, which is a service this install calls rather than a name it goes by.

One judgement call worth flagging: `DEMO_WRITE_ERROR` is a module-level `const` read from fourteen call sites across six controllers. Turning it into a function to reach one word would have spread churn through files this PR otherwise does not touch, and "Self-host TREK for full functionality" is arguably about the project rather than the instance anyway, so it is left as it is.

A renamed install sends email without TREK's logo and the "Travel Resource & Exploration Kit" tagline, which is what [TRADEMARKS.md](https://github.com/liketrek/TREK/blob/main/TRADEMARKS.md) asks of anything not called TREK.

Addresses discussion #2353.

It is a maintenance refactor rather than a feature — the behaviour of every existing install is unchanged — so the diff is here to look at alongside the proposal. Happy to close it if the direction is not wanted, or to reshape the scope.

## Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [x] Documentation update

## Test plan

- `server`: full suite green — 533 files, 11 123 tests, rebased onto `dev` at 3ab5ddaa. `shared`: 58 files, 584 tests. `npm run typecheck` and `npm run typecheck:tests` clean, `npm run lint:check` with 0 errors.
- New `server/tests/unit/nest/app-name.surfaces.test.ts` (APPNAME-001…011) walks every surface in both directions — unset asserts the exact legacy bytes, set asserts the new ones — including the RFC 5545 escaping of a name containing `;` and `,`.
- New cases in `mailer.service.test.ts` (MAILER-015…023): subject, `From` object form, a sender that already carries a display name, `MAIL_FROM_NAME` alone, the password-reset mail in both its text and HTML parts, the admin test send, and an `APP_NAME` full of HTML metacharacters that must not escape the header.
- `derive.test.ts` pins `APP_NAME` in `deriveApp` and `MAIL_FROM_NAME` in `deriveSmtp`, including blank and whitespace-only falling back.
- `calendar.service.test.ts` CAL-020b covers `PRODID`, `X-WR-CALNAME` and the filename for an untitled trip; the pre-existing CAL-020 still asserts `trek-trip.ics` unchanged.
- `maps.service.test.ts` MAPS-096b covers the renamed `User-Agent` still carrying the upstream link.
- `nest-mcp/attach.test.ts` covers the `describe` hook end to end through a real MCP client, and that descriptions are untouched when no transform is configured.
- Three tests guard the placeholder seam, because a `{appName}` that reaches a reader is worse than the stale wording it replaces: `i18n-external-appname.spec.ts` (every locale keeps the placeholder where `en` has one, none invents another, plus a canary so the set can never empty silently), MAILER-022b (nothing survives into a rendered mail) and APPNAME-014 (nothing survives any event text, in any locale). Verified by mutation — breaking `getEventText` fails three of them.
- APPNAME-012/013 pin `withAppName` itself: every occurrence substituted, text without the placeholder returned unchanged (a locale that still spells the name out renders stale wording rather than breaking), and the live name read when none is passed.

## Checklist


mauriceboe:
Appreciate the care that went into this, especially keeping the default output byte-identical. We're not going to take it though.

Running TREK under a different name isn't something we want to support in core. The name is part of the project's identity, and an instance that presents itself as something else is effectively its own distribution. That kind of change belongs in a fork rather than behind a variable upstream.

If you do go the fork route, keep the license in mind. TREK is AGPL-3.0, so a modified version, a renamed one included, that people use over a network has to offer those users its complete source, changes and all, under the same license (section 13 of LICENSE).

It wouldn't make maintenance lighter for us either. Every MCP description and notification that names the app would have to carry the `{appName}` placeholder from here on, nest-mcp gains a transform hook just for that, and renaming an instance later leaves passkeys and authenticator entries under two different names.

For next time, please pitch in `#github-pr` and wait for a go before writing code, as CONTRIBUTING.md asks. This PR was opened before its discussion existed. Closing this and #2353.

PR #2373, opened 2026-09-15
fix(journey): a place kept across two days is two days of the journal

### LINKED ISSUE #2329: [BUG] Journey doesn't show duplicated places
### Pre-flight checklist

### TREK version

4.2.1

### Describe the bug

I tend to set my trip plan so that the same place is used as the end of one and beginning of the following day.

That works fine in trip view but now, filling up Journey I found the places are deduplicated. 

I can give countless examples where this is problematic - you land in a city in the evening, snap a pic of the sunset and the next day morning you have a short sightseeing with more pictures. If you only have a single place then you need to include all photos (evening and morning shots) the day before. 

### Steps to reproduce

1. Create a trip of few days
2. add places
3. assign the same place to two days back to back
4. go and create journey
5. notice that this place is listed only for the first day it was assigned to

### Expected behavior

I expected to have the duplicated days appear in journey

### Deployment method

Docker Compose

### Host OS

Synology DSM

### Accessing TREK from

Desktop browser

### Browser (if applicable)

Brave latest

```shell

```

### Screenshots

Image

hard to snap proper screenshot on an already filled journey - but the same place is not appearing on a next day.

Image

### Additional context

_No response_

subdee:
## Description
`JourneyDomainService` keyed a journal skeleton by `source_place_id` alone. The sync
query at `server/src/nest/journey/journey-domain.service.ts:495` joins places against
`day_assignments`, so a place standing on two days yields two rows — but the dedup set
at `:509`, and `reconcileTripSkeletons`' explicit `placeById` map, collapsed them to
one. `onPlaceCreated` read a single assignment with `.get()`, and `onPlaceUpdated`
applied one arbitrary assignment's date to every entry of that place. The result: a
stop kept across two days got an entry on the first and nothing on the second, so the
next morning's photographs had to go on the previous evening.

The fix adds `journey_entries.source_assignment_id` and keys every sync path on
(place, assignment). The assignment row rather than the day, because `moveAssignment`
moves a stop with `UPDATE day_assignments SET day_id = ?` keeping the row id — keyed
this way an entry still follows a day move instead of being read as one stop leaving
the plan and another arriving. The column deliberately carries no `REFERENCES` clause:
unassigning has to leave the id stale, because a key that no longer matches is exactly
how reconciliation recognises a stop that left; `day_assignments.id` is AUTOINCREMENT,
so it is never handed out twice. Existing rows are backfilled to the place's earliest
assignment — the one the old engine would have picked — and reconciliation adopts a row
that has no link yet rather than detaching and annotating it out from under its author.

This is the sync engine's layer: the skeleton model is derived server-side from trip
assignments, the shared contract does not model entry provenance, and the client groups
entries by date with no place-keying, so nothing on either side of the wire changes shape.

Closes #2329

## Type of Change
- [x] Bug fix

## Checklist


mauriceboe (changes requested):
@subdee test is red

PR #2391, opened 2026-09-18
feat(documents): two-way sync with Paperless, Papra, Nextcloud, OpenCloud and Synology

mauriceboe:
## Description

A trip can be bound to a folder, tag or space in a self-hosted document store, and documents
then move in both directions: what someone uploads in TREK appears there, and what someone
files there appears in TREK. Five products are supported through four adapters: Paperless-ngx,
Papra, Nextcloud and OpenCloud (one WebDAV adapter), and Synology Drive.

Three decisions shape everything else.

**The connection belongs to the trip, not to a person.** Every other integration here is per
user, and photos are shared only through an opt-in flag. Neither can deliver "everyone on the
trip sees the same documents": visibility would depend on whose credentials fetched a file, and
a member without their own Paperless account would see nothing. So the trip owner binds the trip
once, the server talks to the provider under that one identity, and TREK's own membership
decides who sees what. The provider never learns that TREK has members. When the credential
holder leaves the trip the binding goes to `orphaned` rather than quietly carrying on with an
ex-member's token.

**TREK keeps the bytes.** A synced document is an ordinary `trip_files` row, so download, trash,
offline cache, PDF export and backup keep working untouched, and a provider outage does not
blank out a trip that starts tomorrow.

**Deletion never crosses the boundary on its own.** A document that disappears upstream is
recorded as missing and waits for a human. A run that sees more than 40% of its known documents
vanish at once refuses the whole listing rather than acting on it: an unmounted share answers
exactly like a mass delete.

### How it reconciles

The core enumerates both sides on every run rather than asking for a delta, because the weakest
provider sets the model. A changed-since query cannot see a deletion at all; Papra does not move
`updatedAt` when a tag changes (measured against a live instance, not assumed); a Paperless bulk
edit leaves `modified` alone; and Synology FileStation has no change feed whatsoever. A webhook
is therefore only ever "look now", never a source of truth.

Echo is caught by comparing content, not timestamps: `pushed_sha256` records what TREK itself
wrote, so a provider reporting TREK's own upload back is not mistaken for a stranger's edit. A
time window would misfire on any clock skew between TREK and a NAS.

Which side renamed is decided against the name both sides last agreed on, not against the
binding's direction. Both sides renamed goes to a person rather than to a heuristic.

### Per provider

| | Scope | Push | Stable id | Content hash in listing |
|---|---|---|---|---|
| Paperless-ngx | tag + `trek_trip_id` custom field | self-registered workflow webhook | document id | yes |
| Papra | organisation + tag | manual (webhook CRUD is session-only) | document id | yes (sha256) |
| Nextcloud | folder, anchored on `oc:fileid` | `webhook_listeners`, admin only | yes | no |
| OpenCloud | space, anchored on `driveId` | none | yes | no |
| Synology Drive | path under a shared folder | none, polling | no, path is the identity (renames are followed by size and modification time) | no |

Capability differences are answered by data, never by branching on a provider id.

### Admin and setup

Providers are switched on under Admin → Addons → Documents, as shelf rows under the Documents
tile on the desktop and the phone, mirroring how photo providers hang off Journey, including the
cascade when the parent addon is switched off and the 409 when enabling one underneath a
disabled parent. Switching a provider off pauses its bindings without touching them, and they
resume where they stopped once it is back on. The per-trip setup lives in the file manager,
where the person has the context that makes the folder choice obvious; the sync button shows on
a bound trip, or to someone who may bind it while a provider is on. Only the trip owner or an
instance admin can change a binding; every member can see where their documents go.

### Verified against live instances

Each adapter was driven against the real product in Docker (Paperless-ngx 3.1.3, Papra 26.6.2,
Nextcloud 31.0.14, OpenCloud 8.0.1), and Synology against a FileStation stub that reproduces the
documented API including its quirks. DSM cannot be containerised and no NAS was available, so
that adapter is **not** verified against real hardware.

The acceptance run is 339 checks in four suites, all green: connect, scope creation, push, pull,
the echo guard, renames in both directions, deletion on either side under both policies, restore
from TREK's trash in its three variants, purge, the mass-delete guard, members, direction locks,
the scheduler and the webhooks, the admin switch, and Synology two-factor across a server
restart, for all five providers wherever they apply.
Renaming in TREK is reported as not applicable because TREK has no rename for documents;
`FilesService.updateFile` takes no `original_name`. The sync half of it is implemented and
unit-tested and becomes reachable the moment the file manager grows one.

Three bugs the live run found and this PR fixes: an `ON CONFLICT` that did not name the partial
index's predicate and so failed every run; renames resolving in TREK's favour on a `both`
binding, silently renaming a tidied folder back; and the webhook route missing from the
boot-time public-route allow-list. The second run found a cursor that hid a deletion upstream
right after a push; that is fixed as well.

### Also in this PR

`98e293803` fixes uploaded place images not filling their map marker, reported on Discord. It
has nothing to do with documents and is kept as its own commit.

Addresses discussion #214 (integration requests: Paperless-ngx, Papra, Synology Drive,
OpenCloud/Nextcloud). Collabora is deliberately out of scope.

## Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update

## Checklist


mauriceboe:
## Still open after the audit pass

The audit turned up 45 findings. 40 are fixed and pushed (latest `9f531d1cd`). These five are not, with what I know about each so far. Numbers 1 and 2 should land before this ships; 3 to 5 can wait.

Green as of that commit: 533 server tests in `tests/unit/nest/doc-sync`, 265 client tests across `src/components/Files/docsync` and `src/mobile`, both typechecks, i18n parity over all 23 locales.

### 1. `local_deleted` is re-planned on every run, so a delete-policy change acts retroactively

`server/src/nest/doc-sync/doc-sync.helpers.ts:276` and `:319`, executed at `doc-sync.service.ts:279`.

Both planner branches emit `local_deleted` as soon as the trip file carries a `deleted_at` and the item has a `remote_id`. There is no guard on `it.state !== 'local_deleted'`, so the action is planned again on every run for as long as the soft-deleted row sits in `trip_files`.

With `delete_policy = 'unlink'` nothing visible happens. Switch the binding to `trash` later and the next run deletes **every** document at the provider that was ever deleted in TREK, including ones from months ago. With `trash` from the start, `provider.trash()` is called again on every run: idempotent, but pointless load and noise in the log.

Repro: bind with `unlink`, delete a file in TREK, run once (the item goes to `local_deleted`), then switch the binding to "Move to recycle bin" and run again. The document disappears at the provider.

Fix: skip both branches when `it.state === 'local_deleted'`. The deletion then takes effect exactly once, under the policy that was in force when the file was deleted. `SyncItemState` already carries `state`. Wants a planner test plus one in `doc-sync.service.test.ts` that walks the policy change.

### 2. Restoring a file from the trash: what happens to its item is unverified

`server/src/nest/files/files.service.ts:327` only sets `deleted_at = NULL`; caller is `files.controller.ts:261`.

The sync item stays on `local_deleted`. What the next run makes of that is **not verified**. Probably `touch` (bytes unchanged), or `mark_remote_missing` if the provider copy is already gone because the policy was `trash`.

Needs reproducing before anything is changed. Case A, provider copy still there: the item should go back to `synced`. Case B, provider copy trashed: the file has to be uploaded again rather than sitting on `remote_missing` forever. This is the one with the most guesswork in it, so no patch until it has been run.

### 3. Synology device-token cache is keyed by account, not by connection

`server/src/nest/doc-sync/providers/synology-drive.client.ts:474`, key built in `accountKey()` at `:483` from `baseUrl` plus `username`.

Two TREK connections to the same NAS with the same user share the token. The comment above it argues for exactly that: the token belongs to the account and the device, not to the password, and has to survive a password edit and the single-use OTP being cleared.

Most likely deliberate, so I left it alone. What is worth checking is whether two trips with different owners against the same NAS account can pull the token out from under each other; if they can, `connection_id` belongs in the key. Only testable against a real NAS, and the test lab runs Synology against a stub.

### 4. Transport code is duplicated across the four adapters

`paperless.client.ts`, `papra.client.ts`, `webdav.client.ts`, `synology-drive.client.ts` each carry their own timeout `AbortSignal`, size cap, error mapping and `safeFetch` call, in slightly different shapes.

Code quality only, no misbehaviour: acceptance is green across all five providers. This is also the change most likely to break something, and some of the differences are intentional (Papra needs its own error codes, WebDAV its own headers). Separate PR after the release, not this one.

### 5. Cosmetic

- `snapshotHash` in `document-provider.ts` is never read: drop it or use it.
- `connectionId` sits in the scope schema although the caller already has it from the route.
- `createLink` takes a `scope` argument that is always `null`.
- `settings_key` / `payload_key` in `document_provider_fields` are never read.
- Provider class names are inconsistent (`PapraProvider` vs `SynologyDriveProvider` vs `WebdavProvider`).
- `isTripOwner` is named for ownership but checks "may manage": rename it (`canManageDocSync`, as on the client) or fix the comment.

